PR #11, opened 2018-02-15
Setup additional public and private subnet

zaro0508:
Add two additional public and private subnet to the vpc.

diagram: https://imgur.com/a/FH1Mm

### DIFF
--- a/cf_templates/vpc.yml
+++ b/cf_templates/vpc.yml
@@ -475,7 +475,7 @@ Outputs:
       Ref: "VPC"
     Export:
       Name:
-        !Join ['-', [!Ref 'AWS::Region', !Sub '${AWS::StackName}', 'VPC']]
+        !Join ['-', [!Ref 'AWS::Region', !Sub '${AWS::StackName}', 'VPCId']]
   PublicSubnet:
     Description: "SubnetId of the public subnet"
     Value: !Ref PublicSubnetAz0
@@ -512,9 +512,9 @@ Outputs:
     Export:
       Name:
         !Join ['-', [!Ref 'AWS::Region', !Sub '${AWS::StackName}', 'PrivateSubnet2']]
-  DefaultSecurityGroup:
-    Description: "DefaultSecurityGroup Id "
+  VpcDefaultSecurityGroup:
+    Description: "VPC DefaultSecurityGroup Id "
     Value: { "Fn::GetAtt" : ["VPC", "DefaultSecurityGroup"] }
     Export:
       Name:
-        !Join ['-', [!Ref 'AWS::Region', !Sub '${AWS::StackName}', 'DefaultSecurityGroup']]
+        !Join ['-', [!Ref 'AWS::Region', !Sub '${AWS::StackName}', 'VpcDefaultSecurityGroup']]
--- a/cf_templates/vpc.yml
+++ b/cf_templates/vpc.yml
@@ -1,7 +1,6 @@
 AWSTemplateFormatVersion: "2010-09-09"
 Description: >-
-  Creates a VPC with Managed NAT, similar to the VPC Wizard at
-  https://console.aws.amazon.com/vpc/home#wizardFullpagePublicAndPrivate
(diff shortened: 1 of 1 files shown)


zaro0508:
We segment 4094 hosts per subnet so theoretically we can setup up to 16 subnets.   There's no reason for 6 subnets except that it's probably all we need now.  We can add subnets later on. 

There's also no reasoning for AZs but I think you're asking why I've put  "PublicSubnetAz0" and "PublicSubnet1Az0" subnets in the same AZ?  I'm not even sure we need 3 public and 3 private subnets.  I was thinking about removing the "PublicSubnetAz0" and "PrivateSubnetAz0"  subnets. 

Let me know your preference on these.

xschildw:
My point was more that we should be able to specify the AZs (i.e. not forced to take 0 and 3 in the list). I agree that 2 public + 2 private should be enough (1 public + 1 private / AZ).

cf_templates/vpc.yml
```diff
@@ -81,6 +88,96 @@ Resources:
...
+  PublicSubnet1Az0:
+    Type: "AWS::EC2::Subnet"
+    Properties:
+      MapPublicIpOnLaunch: true
+      VpcId: !Ref VPC
+      CidrBlock: !Join
+        - '.'
+        - - !Ref VpcSubnetPrefix
+          - !FindInMap [SubnetConfig, Public1, CIDR]
+      AvailabilityZone: !Select
```
xschildw: Can we pass the index as a param?

cf_templates/vpc.yml
```diff
@@ -81,6 +88,96 @@ Resources:
...
+    Type: "AWS::EC2::Subnet"
+    Properties:
+      MapPublicIpOnLaunch: true
+      VpcId: !Ref VPC
+      CidrBlock: !Join
+        - '.'
+        - - !Ref VpcSubnetPrefix
+          - !FindInMap [SubnetConfig, Public2, CIDR]
+      AvailabilityZone: !Select
+        - 3
```
xschildw: Same here

cf_templates/vpc.yml
```diff
@@ -81,6 +88,96 @@ Resources:
...
+  PrivateSubnet1Az0:
+    Type: "AWS::EC2::Subnet"
+    Properties:
+      VpcId: !Ref VPC
+      CidrBlock: !Join
+        - '.'
+        - - !Ref VpcSubnetPrefix
+          - !FindInMap [SubnetConfig, Private1, CIDR]
+      AvailabilityZone: !Select
+        - 0
```
xschildw: Same

cf_templates/vpc.yml
```diff
@@ -81,6 +88,96 @@ Resources:
...
+  PrivateSubnet2Az3:
+    Type: "AWS::EC2::Subnet"
+    Properties:
+      VpcId: !Ref VPC
+      CidrBlock: !Join
+        - '.'
+        - - !Ref VpcSubnetPrefix
+          - !FindInMap [SubnetConfig, Private2, CIDR]
+      AvailabilityZone: !Select
+        - 3
```
xschildw: Same

xschildw:
What is the use for PublicSubnetAz0 and PrivateSubnetAz3? They are setup differently than the other subnets (where public and private are in the same AZ). My request is basically to replace the hardcoded indices 0 and 3 by 2 params 'zone1' and 'zone2' so we can decide in which zones to deploy the subnets.

zaro0508:
Your request is doable.  However we might want to give those parameters more specific names but I guess it would depend on how we want to organize those zones.  for example, do we want to put all public subnets in zone1 and all private subnets in zone2?  then maybe we should call it "publiczone" and "privatezone"?  The alternative is to put each subnet in different zones, in that case we may want to pass in publiczone1, publiczone2, privatezone1, and privatezone2?

xschildw:
Khai, after discussion I understand the change. Please go ahead and pull it in. Thanks!

PR #16, opened 2018-02-23
Setup VPN security group

zaro0508:
Add security group to allow VPN traffic from the sophos VPN. This
allows the sophos VPN to give users access to AWS resources.

To give VPN users access to an AWS resouce this security group
must be added to that resource.


zaro0508:
Yes, this duplicates Erik's configurations in AWS only.  The only manual steps left are configuring the  sophos web admin.

PR #34, opened 2018-05-14
Peer a single route

xschildw:
Same as peer-route-config but for single route (for cases of subnets that are not paired).


zaro0508:
This looks good however you'll need to rebase.  I renamed the "cf_templates" folder to just "templates" and all CF templates should go in there: https://github.com/Sage-Bionetworks/aws-infra/pull/35

PR #51, opened 2018-07-03
Add template for general notifications

zaro0508:
Add a template to create general notifications for each AWS account.
The notifications are published to SMS, Email, and Slack channel.
Setting up SMS and slack notifications are optional.

The idea is to setup these general SNS topis then have alarms
for various services publish to these topics.  These will serve
as generic notifications for any aws resource.


zaro0508:
Yes, that's the idea.

PR #54, opened 2018-07-18
Setup KMS key on bootstrap

zaro0508:
One of the first things that needs to be done when setting up
CI for an AWS account is to add secrets to the SSM.  To add
secrets to the SSM you need a KMS key.  We add a KMS key to
bootstrap so we can add secrets and run sceptre to retrieve
them and inject them into CF templates.

The idea is to move the infra KMS key from the `essentials`
stack to the `bootstrap` stack.  The one in essentials is
deprecated.


zaro0508:
Decided that this is not necessary

PR #74, opened 2018-12-13
update doc with more info

zaro0508:


zaro0508:
Much of the documentation required requires cross account references therefore I decided that it's better to doc in our wiki.  I have added the docs to our Setup AWS accounts wiki page therefore this PR is no longer needed.

PR #91, opened 2019-03-01
[IT-488] share templates

zaro0508:
This is a DRY refactor.  The SynapseExternalBucket template is
duplicated in multiple repos. Move it to this repo so it can be shared
by sandox-provisioner and sciomp-provisioner repos.  We also rename
it because this script can actually provisions generic buckets as well
as synapse specific buckets.


zaro0508:
yes, but will do that on follow up PRs

PR #94, opened 2019-03-05
Move templates to aws-infra.

xschildw:
Moving 2 templates to aws-infra so they can be shared.

### DIFF
--- a/ci/taskcat.yaml
+++ b/ci/taskcat.yaml
@@ -56,14 +56,16 @@ tests:
     regions:
       - us-east-1
     template_file: s3-bucket.yaml
-  s3web-bucket:
-    parameter_input: s3web-bucket-input.json
-    regions:
-      - us-east-1
-    template_file: managed-s3Web.yaml
-  s3web-cloudfront-bucket:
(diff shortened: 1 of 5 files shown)


xschildw:
Created ACM in admincentral, added tests and input configs. Looks OK?

ci/taskcat.yaml
```diff
@@ -56,3 +56,14 @@ tests:
     regions:
       - us-east-1
     template_file: s3-bucket.yaml
+  s3web-bucket:
```
zaro0508: This looks good.  I just tried to run it and realized that I forgot how long cloudfront takes to deploy.  These adds about 60 mins to the tests.  Not only that, the stack deletion fails when taskcat attempts to delete buckets.  Let's just comment out these for now with the following note:

For s3web-bucket:
"Stack deletion fails when taskcat attempts to delete bucket"

For s3web-cloudfront-bucket:
"Cloudfront adds to much time to the test (~60 mins to deploy and cleanup).  Also taskcat fails to cleanup."

xschildw: OK. Thanks for the feedback!

PR #98, opened 2019-03-15
Add Lambda to create AWS AMIs

zaro0508:
Add a lambda that we can use to provision AWS AMIs.  Also deploy it
to our lambda repository on admincentral.


thomasyu888:
How does one use this tool?  PR with the stack name or url?

zaro0508:
ok, this was a bad idea.  a better idea is to leverage aws launch templates to create AMIs.

PR #100, opened 2019-03-27
[IT-520] Setup role for DLM

zaro0508:
Setup a role for the Data lifecycle manager[1]

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/snapshot-lifecycle.html


zaro0508:
yes, that is correct.

PR #105, opened 2019-04-02
[IT-529] Update synapse bucket permissions

zaro0508:
As explained in these updated instructions:
https://github.com/Sage-Bionetworks/synapseDocs/pull/538/files

When an S3 bucket is to be used with Synapse it needs to grant Synapse
`S3:GetBucketLocation` permission, so Synapse can determine the Region
and configure the S3 client accordingly.


zaro0508:
@brucehoff should we merge this yet?

PR #108, opened 2019-04-11
Allow user to use specific AMI

zaro0508:
Allow user to setup a batch job with a custom AMI.  This is an
optional parameter.  If you does not specify an AMI then batch
will use a default AWS AMI.

### DIFF
--- a/templates/managed-batch.yaml
+++ b/templates/managed-batch.yaml
@@ -35,8 +35,13 @@ Parameters:
     Description: The schedule (cron) expression to trigger the batch job
     Type: String
     Default: "rate(1 day)"
+  AMIId:
+    Description: ID of the AMI to deploy
+    Type: String
+    Default: ""
 Conditions:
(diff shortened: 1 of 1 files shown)


templates/managed-batch.yaml
```diff
@@ -168,6 +173,7 @@ Resources:
         MinvCpus: 0
         DesiredvCpus: 0
         MaxvCpus: 64
+        ImageId: !If [HasAMIId, !Ref AMIId, !Ref "AWS::NoValue"]
```
thomasyu888: Do you know if the cloud formation template will work if there is no value?

zaro0508: i have not tested but we do have a test that will tell us. https://github.com/Sage-Bionetworks/aws-infra/blob/master/ci/managed-batch-input.json

the test does not set this param so if the test passed then we know it will work.

PR #111, opened 2019-04-15
update defaults

zaro0508:
* remove regions that are not supported
* make keyname an optional parameter. if nothing specified no SSH key pair will
be assigned to the instance.

### DIFF
--- a/templates/managed-ec2-v7.yaml
+++ b/templates/managed-ec2-v7.yaml
@@ -6,7 +6,7 @@ Parameters:
     Description: Name of an existing EC2 KeyPair to enable SSH access to the instance
     Type: 'AWS::EC2::KeyPair::KeyName'
     ConstraintDescription: must be the name of an existing EC2 KeyPair.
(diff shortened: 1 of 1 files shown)


templates/managed-ec2-v7.yaml
```diff
@@ -495,7 +440,7 @@ Resources:
     Properties:
       ImageId: !If [HasAMIId, !Ref AMIId, !FindInMap [AWSRegionArch2AMI, !Ref 'AWS::Region', !FindInMap [AWSInstanceType2Arch, !Ref InstanceType, Arch]]]
       InstanceType: !Ref InstanceType
-      KeyName: !Ref KeyName
+      KeyName: !If [HasKeyName, !Ref KeyName, !Ref 'AWS::NoValue']
```
xschildw: May have been asked before, if no value then the instance is only accessible from JumpCloud account?

zaro0508: that is correct

PR #121, opened 2019-05-01
[IT-408] support ubuntu images

zaro0508:
Provision an EC2 instance based off of a AWS ubuntu AMI.

differences between ubuntu template and aws linux template:
* package manager is apt instead of yum
* volume name must be "/dev/sda1" instead of "/dev/xvda"
* cfn tools come preinstalled on aws linux AMI, we need to
  in stall on other distros.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,8 +19,8 @@ jobs:
   include:
     - stage: validate
       script: 
-        - cfn-lint ./templates/**/*.yaml
-        - cfn-lint ./lambdas/**/*.yaml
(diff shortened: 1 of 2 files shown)


zaro0508:
Added an ignore check to cfn-lint for the following check:
```
W1020 Fn::Sub isn't needed because there are no variables at Resources/Ec2Instance/Metadata/AWS::CloudFormation::Init/configure_cfn/files//lib/systemd/system/cfn-hup.service/content/Fn::Sub
templates/managed-ec2-ubuntu-v1.yaml:427:15
```

PR #124, opened 2019-05-08
[IT-567] Allow opening up multiple ports on EC2

zaro0508:
Refactor to allow provisioner to open multiple incoming ports
on an EC2 instance.  Due to the limitations with CFN we need
to convert the CFN templates to jinja[1] template to provide
this funcionality.

Example usage:

provision ec2 and open port 80, 1099, and 8020 to incoming traffic.
```
template_path: remote/managed-ec2-v8.j2
stack_name: test-ec2
sceptre_user_data:
  OpenPorts:
    - "80"
    - "1099"
    - "8020"
```

[1] http://jinja.pocoo.org/docs/2.10/templates/


zaro0508:
Look at the example in the commit message.  It's in the sceptre config file.

PR #131, opened 2019-05-17
refactor managed-ec2-v9.j2 to be cleaner

zaro0508:


zaro0508:
replaced with PR #132

PR #141, opened 2019-05-31
[IT-585] Fix jumpcloud lambda

zaro0508:
The jumpcloud lamba to clean up jumpcloud systems was broken. The
problem was the parameter set to the systems_list method was invalid.
This lambda has also been refactored to a serverless lambda for ease of
maintenance.


zaro0508:
No.  The purpose is to remove the hosts from the jumpcloud inventory when it gets removed from the provisioner's inventory (or AWS).

zaro0508:
hmm, I just noticed that jumpcloud added filtering to their inventory so now we can just filter out inactive systems on the jumpcloud UI.  Not sure we need this anymore but probably will commit to git and just not use for now.

PR #150, opened 2019-06-13
[IT-583] Encrypt volume launch

zaro0508:
* AWS now supports deploying encrypted AMI from non-encrypted AMI[1].
  Add a parameter to allow users to provision an instance with an
  encrypted root volume
* Occasionally googlechrome will fail to instance from chocolately repo
  due to not having an updated checksum.  Use '--ignore-checksums' flag
  so make it less error prone[2].  Not this is not as secure but trade
  of is to make it less error prone.
* New set of version files because this change may cause AWS to replace
  instances if updating existing stacks.

[1] https://aws.amazon.com/blogs/security/how-to-quickly-launch-encrypted-ebs-backed-ec2-instances-from-unencrypted-amis/
[2] https://www.gep13.co.uk/blog/chocolatey-error-hashes-do-not-match

### DIFF
--- a/templates/managed-ec2-ubuntu-v3.j2
+++ b/templates/managed-ec2-ubuntu-v3.j2
@@ -1,0 +1,305 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  Provision EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
+Parameters:
+  KeyName:
+    Description: Name of an existing EC2 KeyPair to enable SSH access to the instance
+    Type: 'AWS::EC2::KeyPair::KeyName'
+    ConstraintDescription: must be the name of an existing EC2 KeyPair.
+    Default: ""
+  InstanceType:
+    Description: WebServer EC2 instance type
+    Type: String
+    Default: t3.nano
+  JcConnectKey:
+    Description: Connection key used to let JC agent connect to a Jumpcloud account.
+    Type: String
+    NoEcho: true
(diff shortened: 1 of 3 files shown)


templates/managed-ec2-ubuntu-v3.j2
```diff
@@ -0,0 +1,305 @@
...
+    Type: String
+    AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
+    ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
+  EncryptVolume:
+    Type: String
+    Description: true to encrypt root volume, false (default) for no encryption
+    AllowedValues:
+      - true
+      - false
+    Default: false
```
thomasyu888: So default is false? Does this mean that people on scicomp have to set this to true? I don't like that.. for security, maybe default should be true?

thomasyu888: Or this could just be specified in the PR template. Saying that it **has to** be true in scicomp.

zaro0508: My idea was to set it to true by default in the PR template.

PR #152, opened 2019-07-02
[IT-305] Setup dead letter queue for lambda

zaro0508:
Setup deadletter queue[1] for SAM based lambda to send notifications to

[1] https://github.com/awslabs/serverless-application-model/blob/master/versions/2016-10-31.md#deadletterqueue-object


zaro0508:
Good question.  I'm not sure but i'm guessing that if we attempt to send to an invalid email then it should fail delivery.

PR #154, opened 2019-07-09
[IT-557] one template for aws and ubuntu linux

zaro0508:
The automated provisioner currently supports two linux distributions
(aws linux and ubuntu).  We manage them with 2 separate cloudformation
templates (managed-ec2-*.yaml and managed-ec2-ubuntu-*.yaml).  This
change creates one template that supports provisioning both distros.
This will make the sage provisioner easier to maintain.

To set the distro add the following to the sceptre template:

```
sceptre_user_data:
  Distro: "ubuntu"     # AMIId must match distro
parameters:
  AMIId: "ami-0a313d6098716f372"    # Ubuntu, 18.04 LTS, amd64 bionic image build on 2019-02-12
```

### DIFF
--- a/templates/managed-ec2-linux-v1.j2
+++ b/templates/managed-ec2-linux-v1.j2
@@ -1,0 +1,324 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  Provision EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
+Parameters:
+  KeyName:
+    Description: Name of an existing EC2 KeyPair to enable SSH access to the instance
+    Type: 'AWS::EC2::KeyPair::KeyName'
+    ConstraintDescription: must be the name of an existing EC2 KeyPair.
+    Default: ""
+  InstanceType:
+    Description: WebServer EC2 instance type
+    Type: String
+    Default: t3.nano
+  JcConnectKey:
+    Description: Connection key used to let JC agent connect to a Jumpcloud account.
+    Type: String
+    NoEcho: true
+  JcServiceApiKey:
+    Description: The Jumpcloud service user API key
+    Type: String
+    NoEcho: true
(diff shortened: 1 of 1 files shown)


zaro0508:
Maybe? Hardening kit is orthogonal to this change.  Please create a jira issue and we can discuss that separately.

thomasyu888:
Hmm. What would an example PR look like from people? Would people still need to specify the AMI Ids

zaro0508:
The PR would look exactly the same except there would be one additional parameter `Distro` and the `AMIId` must match the distro.  Take a look at the commit message.

thomasyu888:
I see. So you would have to:

```
sceptre_user_data:
  Distro: "ubuntu"     # AMIId must match distro
  # Distro: "linux"     # AMIId must match distro
parameters:
  AMIId: "ami-0a313d6098716f372"    # Ubuntu, 18.04 LTS, amd64 bionic image build on 2019-02
  # AMIId: "ami-0b898040803850657"      # Amazon Linux 2 AMI 2.0.20190618 x86_64 HVM gp2
```

I mean seems fine to me, unsure if this would really help users.  At the end of the day, I still have to go on the github page to look for the newest github template, then copy and paste into a file.  So if I have to do that already, I can just easily pick ubuntu or linux template.

zaro0508:
Correct.  The benefit is that users would only need to find a  one `linux` example template and not need to find a file for a specific distro.  The bigger benefit is on maintaining these templates particularly if we need to support other distros like redhat, SUSE, centos, etc..

thomasyu888:
Either way, users are going to have to learn to use the new template.  I'm indifferent about this change, but I can see how it would be helpful for IT in the long run.

zaro0508:
Merging doesn't mean we are going to use it but it will facilitate further testing.

PR #164, opened 2019-07-22
increase volume size limit

zaro0508:
Some people have requested instances with more than 1TB disk.


thomasyu888:
Will this end up updating people's stacks?

zaro0508:
@ahayden prevents people from requesting ridiculously large volumes.
@thomasyu888 no, it's just a limit, not the value of the actual disk size.

PR #172, opened 2019-08-23
packer resources

zaro0508:
Packer creats an AMI and boostraps it by connecting to it with SSH and
running scripts (sh/ansible/puppet/etc..) to setup the instance.
Accounts that want to use packer need to setup a security group
to allow this connection.


zaro0508:
This is not needed.  Packer will create a temporary security group when it provisions an AMI.

PR #173, opened 2019-08-29
Whitelist vpc access for external S3 region-restricted buckets (IT-483) (IT-661)

zimingd:


zimingd:
@zaro0508 I removed the paramer change in s3-bucket.yaml

PR #176, opened 2019-09-10
[IT-418] fix cyclical dependency

zaro0508:
Break cyclical dependency in cloudformation template by unmasking
the aws account id.  Also we make it a default so downstream
templates don't won't need to set this parameter at all.


zaro0508:
@ahayden made me realize that all this peering stuff may just dissappear from our templates after we make the move from transit VPC to the transit gateway.  I think that's a better strategy.

PR #177, opened 2019-09-11
[IT-666] remove CI group

zaro0508:
Our CI user, travis, does not need admin permissions.  Our CI system
is already setup to have travis assume the CfService role to deploy
templates.  For better security we revoke permissions from travis.


zaro0508:
it's a CFN template thus only deals with AWS access, not Github.

PR #180, opened 2019-10-02
Added SSM Service Role allow rule

ahayden:
@zaro0508 I think this is all we need to give SSM services access to instances. This is just permissions; we would configure the services separately based on the users' needs for the instance, i.e. maintenance windows, etc.


ahayden:
Pls don't merge until we talk about how it works ;/

ahayden:
Welp concatenating roles does not build.

templates/essentials.yaml
```diff
@@ -429,6 +429,21 @@ Resources:
...
+        Version: '2012-10-17'
+        Statement:
+        - Effect: Allow
+          Principal:
+            Service:
+            - 'ec2.amazonaws.com'
+          Action: 
+            - 'sts:AssumeRole'
+      ManagedPolicyArns: 
+            - 'arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore'
```
zaro0508: From reading the [documentation](https://docs.aws.amazon.com/systems-manager/latest/userguide/automation-cf.html), it looks like you actually want the `arn:aws:iam::aws:policy/service-role/AmazonSSMAutomationRole`  role?

ahayden:
Going to change the approach

templates/essentials.yaml
zaro0508: why not set this to `AmazonSSMAutomationRole`?

you will also need to add the policy from `TagRootVolumeRole` resource here as well.

templates/essentials.yaml
```diff
@@ -429,6 +429,21 @@ Resources:
                   - "ec2:CreateTags"
                 Resource:
                   - 'arn:aws:ec2:*::snapshot/*'
+  AWSSSMManagedInstanceRole:
```
zaro0508: don't we want to create a `ManagedInstanceProfile` and associate this role with the profile?

templates/essentials.yaml
ahayden: I believe the `AmazonSSMAutomationRole` is meant for resources that would make changes to instances that have the `AmazonSSMManagedInstanceCore` policy attached. Like [this example](https://docs.aws.amazon.com/systems-manager/latest/userguide/automation-walk-security-assume.html).

zaro0508: took another look and you are correct.

templates/essentials.yaml
```diff
@@ -456,7 +471,7 @@ Outputs:
     Export:
       Name: !Sub '${AWS::Region}-${AWS::StackName}-InfraKeyAlias'
   TagRootVolumeProfile:
-    Description: Profile to allow instances to tag its root volume
+    Description: Profile to allow instances to adopt roles and interact with services
```
zaro0508: this is not accurate for `TagRootVolumeProfile`

templates/managed-ec2-core.yaml
```diff
@@ -0,0 +1,44 @@
+Description: Allows central management of computing instances; building on essentials
+AWSTemplateFormatVersion: 2010-09-09
+Parameters:
+  MaintenanceWindow:
```
zaro0508: this is not used in this template, why do we need it?

templates/managed-ec2-linux-v2.j2
```diff
@@ -0,0 +1,345 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  Provision EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
+Parameters:
+  MaintenanceWindow:
```
zaro0508: Your intention is to have people set this?  Hmm, it would be hard to imagine that happening.

templates/managed-ec2-core.yaml
```diff
@@ -0,0 +1,44 @@
...
+    Type: String
+    Default: 'Weekly-0400SundayUTC-7'
+Resources:
+  ComputingInstanceProfile:
+    Type: 'AWS::IAM::InstanceProfile'
+    Properties:
+      Path: "/"
+      Roles:
+        - !Ref ComputingInstanceRole
+  ComputingInstanceRole:
```
zaro0508: isn't this already in essentials.yaml? why do you need it in both files?

templates/managed-ec2-linux-v2.j2
```diff
@@ -0,0 +1,345 @@
...
+          GroupSet:
+            - !ImportValue
+              'Fn::Sub': '${AWS::Region}-${VpcName}-VpnSecurityGroup'
+          {% if sceptre_user_data.OpenPorts is defined %}
+            - !GetAtt InstanceSecurityGroup.GroupId
+          {% endif %}
+          SubnetId: !ImportValue
+            'Fn::Sub': '${AWS::Region}-${VpcName}-${VpcSubnet}'
+      IamInstanceProfile: !ImportValue
+        'Fn::Sub': '${AWS::Region}-essentials-ComputingInstanceProfile'
```
zaro0508: which stack should you be referencing?  `essentials` or `managed-ec2-core`?

ahayden: I moved the profiles back to essentials

templates/managed-ec2-core.yaml
ahayden: I wanted it in both initially because I wanted to be able to apply it to currently deployed instances and new ones

templates/managed-ec2-linux-v2.j2
ahayden: I want to give people a few options and for it to default to the best window I can think of when they don't select one.

templates/essentials.yaml
```diff
@@ -38,6 +38,10 @@ Parameters:
     AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
     ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
     Default: 'it@sagebase.org'
+  MaintenanceWindow:
+    Description: 'Acceptable outage intervals for Systems Manager changes'
+    Type: String
+    Default: 'Weekly-0400SundayUTC-7'
```
zaro0508: since there is a default for this param people don't need to set it.  So I'm wondering what is the expectation if people don't set it?  how would we know whether people are really ok with this outage interval?

templates/managed-ec2-linux-v2.j2
```diff
@@ -0,0 +1,339 @@
...
+                  - "/usr/bin/aws ec2 create-tags --region $AWS_REGION --resources $ROOT_DISK_ID "
+                  - "--tags Key=Name,Value=$EC2_INSTANCE_ID-root Key=cloudformation:stack-name,Value=$STACK_NAME "
+                  - "Key=Department,Value=$DEPARTMENT Key=Project,Value=$PROJECT Key=OwnerEmail,Value=$OWNER_EMAIL"
+              env:
+                AWS_REGION: !Ref AWS::Region
+                STACK_NAME: !Ref AWS::StackName
+                DEPARTMENT: !Ref Department
+                PROJECT: !Ref Project
+                OWNER_EMAIL: !Ref OwnerEmail
+      ImageId: !Ref AMIId
```
zaro0508: what happened to `Properties:`?

ahayden: I deleted it! fixing...

templates/essentials.yaml
ahayden: I'm thinking they can choose to opt out if they have a good reason, but it should default to the least invasive maintenance window

zaro0508: I think we need more discussion around this tag.  I'm concerned that it can become unclear whether and not usable.  Also this is a bit unrelated to the goal of this PR. 

To unblock this specific PR I would recommend removing this tag and opening a new PR with only this tag change so we can have a PR dedicated to discussing this new tag.

templates/managed-ec2-win-v5.j2
```diff
@@ -0,0 +1,335 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  Provision EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
+Parameters:
+  MaintenanceWindow:
```
zaro0508: ditto here.

templates/managed-ec2-linux-v2.j2
zaro0508: ditto here.

templates/essentials.yaml
```diff
@@ -465,6 +493,20 @@ Outputs:
...
+    Description: Profile to allow instances to tag and interact with services
+    Value: !Ref ComputingInstanceProfile
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ComputingInstanceProfile'
+  ComputingInstanceRole:
+    Description: Role to allow instances tag and receive management via Systems Manager
+    Value: !Ref ComputingInstanceRole
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ComputingInstanceRole'
+  MaintenanceWindow:
```
zaro0508: ditto here

templates/managed-ec2-win-v5.j2
ahayden: OK, agreed- I updated this PR with a commit that removes this MaintenanceWindow param

PR #181, opened 2019-10-09
setup pre-commit validations

zaro0508:


zaro0508:
It doesn't but this PR includes fixes for all of the broken rules.

PR #182, opened 2019-10-17
template to setup transit gateway attachments

zaro0508:
The workflow for setting up the transit gateway it to provision the
attachments on the spoke accounts of the "hub and spoke" model
therefore this template will be used to setup every hub to
spoke attachement (aka connections).

This depends on PR https://github.com/Sage-Bionetworks/transit-infra/pull/2

### DIFF
--- a/templates/transit-gateway-attachment.yaml
+++ b/templates/transit-gateway-attachment.yaml
@@ -1,0 +1,61 @@
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  Setup an AWS transit gateway attachment
+Parameters:
+  Department:
+    Description: 'The department for this resource'
+    Type: String
+    AllowedPattern: '^\S*$'
+    ConstraintDescription: 'Must be string with no spaces'
+    Default: 'Platform'
+  Project:
+    Description: 'The name of the project that this resource is used for'
+    Type: String
+    AllowedPattern: '^\S*$'
+    ConstraintDescription: 'Must be string with no spaces'
+    Default: 'Infrastructure'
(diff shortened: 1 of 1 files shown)


templates/transit-gateway-attachment.yaml
```diff
@@ -0,0 +1,61 @@
...
+  TransitGatewayAttachment:
+    Type: AWS::EC2::TransitGatewayAttachment
+    Properties:
+      SubnetIds:
+        - !ImportValue
+          'Fn::Sub': '${AWS::Region}-${VpcName}-PrivateSubnet'
+        - !ImportValue
+          'Fn::Sub': '${AWS::Region}-${VpcName}-PrivateSubnet1'
+        - !ImportValue
+          'Fn::Sub': '${AWS::Region}-${VpcName}-PrivateSubnet2'
```
ahayden: Are there any use cases for public subnets? maybe routing to a service through the transit gateway but not through the internet?

zaro0508: Yes.  however AWS attachments only allows a single AZ to single subnet mapping.  We designed our VPCs with multiple subnets in a single zone.   For example, the "PublicSubnet1" and "PrivateSubnet1" both are in "us-east-1a".  This means that we can attach either "PublicSubnet1" OR "PrivateSubnet1" to "us-east-1a" but not both.  I decided that there are more situations where we want to route to private subnets than public ones which is why I've set it up like this. I've never run into this restriction with any other AWS service.  It might be worth looking into refactoring our VPC architecture so that every subnet is in a separate zone.

PR #205, opened 2019-12-19
Add instance types lambda and cfn template

tthyer:
Add a lambda to get all EC2 instance types, then filter on several conditions.


lambdas/instance-types/app/app.py
```diff
@@ -0,0 +1,104 @@
...
+def lambda_handler(event, context):
+  logger.debug('Received event: ' + json.dumps(event, sort_keys=True))
+
+  instance_types = get_instance_types()
+
+  valid_types = filter_types(instance_types)
+
+  yaml = format_types(valid_types)
+
+  write_to_S3(yaml)
```
zaro0508: you don't need to write out to S3, the SAM build will do that for you.  what you want to do is to return the yaml.
```
    return {
        "requestId": event["requestId"],
        "status": status,
        "fragment": fragment,
    }
```

take a look at https://github.com/Sage-Bionetworks/aws-infra/blob/master/lambdas/cfn-explode-macro/explode/app.py#L97

lambdas/instance-types/app/app.py
```diff
@@ -0,0 +1,104 @@
+#! /usr/bin/env python3.7
```
zaro0508: please remove

lambdas/instance-types/template.yaml
```diff
@@ -0,0 +1,136 @@
+AWSTemplateFormatVersion: '2010-09-09'
```
zaro0508: too much stuff in this file.  please follow pattern in  https://github.com/Sage-Bionetworks/aws-infra/blob/master/lambdas/cfn-explode-macro/template.yaml

lambdas/instance-types/app/app.py
```diff
@@ -0,0 +1,104 @@
...
+    'Name': 'bare-metal',
+    'Values': [
+        'false',
+    ]
+  }
+]
+
+# removes the current classes of high memory, accelerated computing,
+# and storage optimized instances
+restricted_prefixes = [
```
zaro0508: [optiona] would be nice to pass this list in as a template parameter

lambdas/instance-types/app/app.py
```diff
@@ -0,0 +1,104 @@
...
+import json
+import logging
+import os
+
+import boto3
+
+logger = logging.getLogger()
+logger.setLevel(logging.INFO)
+
+instance_type_filters = [
```
zaro0508: [optional] would be nice to pass this filter in as a template parameter

tthyer:
closing as it looks like I cannot use an include in parameters

PR #207, opened 2020-01-06
CFN list transform

tthyer:
adding a list transform macro for cloud formation


tthyer:
@zaro0508 Please review

lambdas/cfn-list-transform-macro/list_transform/app.py
```diff
@@ -0,0 +1,32 @@
+import json
+import traceback
+
+
+def handle_transform(parameters):
+  operation = parameters['Operation']
+  transform_string = parameters['TransformString']
+  fragment = []
+  input_list = parameters['List']
```
zaro0508: should this be `parameters['InputList']`?

also maybe you want to check that it's actually a list?

lambdas/cfn-list-transform-macro/list_transform/app.py
```diff
@@ -0,0 +1,32 @@
...
+
+def handler(event, context):
+    status = 'success'
+    response = {
+      'requestId': event['requestId'],
+      'status': 'success'
+    }
+
+    try:
+      response['fragment'] = handle_transform(event['params'])
```
zaro0508: I believe the parameters are in a `templateParameterValues` key, so this should be something like..
```
parameters = event["templateParameterValues"]
fragment = handle_transform(parameters)
```

tthyer:
@zaro0508 Thanks for the feedback. Before I address all these comments, we need to discuss whether this strategy is going to work at all. It appears it won't because one can only use the SecurityGroups in the AWS::EC2::Instance definition if those groups are defined on the default VPC. Since we are mostly using internalvpc and it is not the default, the solution to create all the security group names from the input ports is not going to work, unless you decide to make internalvpc the default vpc. Do you intend to do that?

lambdas/cfn-list-transform-macro/README.md
```diff
@@ -0,0 +1,69 @@
+# cfn-list-transform-macro
```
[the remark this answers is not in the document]
tthyer: added better explanation in 9e5e7cf

lambdas/cfn-list-transform-macro/tests/unit/expected_append.json
```diff
@@ -0,0 +1,33 @@
+{
```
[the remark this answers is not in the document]
tthyer: nope, removing

lambdas/cfn-list-transform-macro/tests/unit/test_handler.py
```diff
@@ -0,0 +1,35 @@
...
+    self.assertEqual(expected, fragment)
+
+
+  def test_missing_prepend(self):
+    with open(r'events/list_transform_append.json') as file:
+      test_template = json.load(file)
+    self.event = test_template
+    result = handler(self.event, None)
+    fragment = result['fragment']
+    expected = ['AlphaFoo', 'BetaFoo', 'GammaFoo']
```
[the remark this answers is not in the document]
tthyer: The point is to confirm that if one does not add the Prepend argument then the behavior is the same as including `Prepend: false`

lambdas/cfn-list-transform-macro/list_transform/app.py
tthyer: No, it should not be `parameters['InputList']` but I will check that it's a list.

tthyer: check and test for check added in b20ab10

lambdas/cfn-list-transform-macro/list_transform/app.py
tthyer: "params" is a sibling of "templateParameterValues" -- [reference](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/template-macros.html#template-macros-lambda-interface)

PR #210, opened 2020-01-15
Add macro for looking up parameters in ssm parameter store

tthyer:
@zaro0508 This is the macro we discussed for pulling values from the parameter store. This PR relies on the template hiding in https://github.com/Sage-Bionetworks/scipoolprod-sc-lib-infra/pull/27.


lambdas/cfn-ssm-param-macro/ssm_param/app.py
```diff
@@ -0,0 +1,35 @@
...
+    if 'Secure' in parameters and parameters['Secure'] is True:
+        secure = True
+    if 'SsmKeyName' not in parameters:
+        raise ValueError('"SsmKeyName" parameter is required')
+    keyname = parameters['SsmKeyName']
+    client = boto3.client('ssm')
+    response = client.get_parameter(
+        Name=keyname,
+        WithDecryption=secure)
+    print(response)
```
zaro0508: this may leak secrets. i suggest removing it.

lambdas/cfn-ssm-param-macro/template.yaml
```diff
@@ -0,0 +1,56 @@
...
+Resources:
+  SsmParamFunction:
+    Type: AWS::Serverless::Function
+    Properties:
+      CodeUri: ssm_param/
+      Handler: app.handler
+      Runtime: python3.6
+      Role: !GetAtt SsmParamExecutionRole.Arn
+      Tags:
+        Department: !Ref Department
```
zaro0508: you'll need to add parameters section for these Refs

lambdas/cfn-ssm-param-macro/template.yaml
```diff
@@ -0,0 +1,56 @@
...
+            Principal:
+              Service: lambda.amazonaws.com
+            Action:
+              - 'sts:AssumeRole'
+      Path: /
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/AmazonSSMReadOnlyAccess
+        - !Ref DecryptSecureKeysPolicy
+
+  DecryptSecureKeysPolicy:
```
zaro0508: would this policy allow anyone the ability to retrieve secrets?  my concern is that anyone can deploy a template and this lambda will execute and expose secrets.  Could we pass in an a principal ARN and restrict this policy to that principal?  We should restrict to `us-east-1-bootstrap-CfServiceRoleArn`

lambdas/cfn-ssm-param-macro/ssm_param/app.py
tthyer: oh dear, good find

lambdas/cfn-ssm-param-macro/template.yaml
tthyer: That is exactly the advice I needed

lambdas/cfn-ssm-param-macro/README.md
```diff
@@ -0,0 +1,100 @@
...
+```yaml
+commands:
+  01_jumpcloud_agent:
+    command: !Join
+      - ''
+      - - "/usr/bin/curl --silent --show-error --header 'x-connect-key: "
+        - Fn::Transform
+          Name: SsmParam
+          Parameters:
+            Secure: true
```
zaro0508: should these be updated to correspond to refactor?

lambdas/cfn-ssm-param-macro/tests/events/string.json
```diff
@@ -0,0 +1,8 @@
+{
+  "fragment": {},
+  "requestId": "foo",
+  "params": {
+    "Type": "String",
+    "Name": "test-param"
```
zaro0508: is there a test where the "Name" key is present however the value does not exist in the SSM?

tthyer: Just added this

lambdas/cfn-ssm-param-macro/template.yaml
```diff
@@ -36,7 +56,8 @@ Resources:
         Statement:
           - Effect: Allow
             Principal:
-              Service: lambda.amazonaws.com
+              AWS:
```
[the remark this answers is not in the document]
tthyer: Implemented an invocation policy distinct from execution policy as discussed in chat

PR #211, opened 2020-01-16
cloudformation macro for S3

zaro0508:
Add a cloudformation macro to help automate provisioning of a Synapse
storage location[1] using the service catalog.  This attempts to resolve
issue https://github.com/Sage-Bionetworks/scipoolprod-sc-lib-infra/issues/19

We got this macro from awslabs[2]

[1] https://docs.synapse.org/articles/custom_storage_location.html
[2] https://github.com/awslabs/aws-cloudformation-templates/tree/master/aws/services/CloudFormation/MacrosExamples/S3Objects

### DIFF
--- a/lambdas/cfn-s3objects-macro/template.yaml
+++ b/lambdas/cfn-s3objects-macro/template.yaml
@@ -1,0 +1,29 @@
+# From https://github.com/awslabs/aws-cloudformation-templates/tree/master/aws/services/CloudFormation/MacrosExamples/S3Objects
+AWSTemplateFormatVersion: '2010-09-09'
+Transform: AWS::Serverless-2016-10-31
+Description: A Cloudformation macro to deploy objects to S3 buckets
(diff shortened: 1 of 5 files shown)


lambdas/cfn-s3objects-macro/template.yaml
```diff
@@ -0,0 +1,29 @@
+# From https://github.com/awslabs/aws-cloudformation-templates/tree/master/aws/services/CloudFormation/MacrosExamples/S3Objects
+AWSTemplateFormatVersion: '2010-09-09'
```
tthyer: @zaro0508 Is this ^^ the only line that is different between the two PRs?

zaro0508: no, i think i made some other changes as well.  I've verified that this is working in sandbox so i think we good to go

PR #214, opened 2020-01-17
Add ssm role for lambda to bootstrap

tthyer:
This is a proposal to move the role used for executing the SsmParam lambda to bootstrap.yaml and export it, then make it one of the principals that can use the AWSKmsInfraKey. Setting the principal on the KeyPolicy should not require an update of the Key itself. That's the first step. 
Step 2 is to set the SsmParam lambda to use this new role.

@zaro0508 what do you think?

### DIFF
--- a/templates/bootstrap.yaml
+++ b/templates/bootstrap.yaml
@@ -106,6 +106,32 @@ Resources:
       Roles:
         -
           !Ref AWSIAMCfServiceRole
+  # This role is used by the SsmParam Lambda to read keys
+  AWSIAMSsmParamLambdaExecutionRole:
+    Type: 'AWS::IAM::Role'
+    Properties:
+      AssumeRolePolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Effect: Allow
+            Principal:
+              Service: lambda.amazonaws.com
+            Action:
+              - 'sts:AssumeRole'
+      Path: /
+      ManagedPolicyArns:
+        - arn:aws:iam::aws:policy/AmazonSSMReadOnlyAccess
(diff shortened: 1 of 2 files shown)


templates/bootstrap.yaml
```diff
@@ -127,3 +153,7 @@ Outputs:
     Value: !GetAtt AWSIAMCfServiceRole.Arn
     Export:
       Name: !Sub '${AWS::Region}-${AWS::StackName}-CfServiceRoleArn'
+  AWSIAMSsmParamLambdaExecutionRole:
+    Value: !GetAtt AWSIAMSsmParamLambdaExecutionRole.Arn
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-SsmParamLambdaExecutionRole'
```
zaro0508: this too

templates/bootstrap.yaml
```diff
@@ -106,6 +106,32 @@ Resources:
       Roles:
         -
           !Ref AWSIAMCfServiceRole
+  # This role is used by the SsmParam Lambda to read keys
```
zaro0508: how about URL reference to the lambda?

templates/essentials.yaml
```diff
@@ -342,6 +342,7 @@ Resources:
               AWS:
                 - !ImportValue us-east-1-bootstrap-TravisUserArn
                 - !ImportValue us-east-1-bootstrap-CfServiceRoleArn
+                - !ImportValue us-east-1-bootstrap-SsmParamLambdaExecutionRole
```
zaro0508: this too

PR #224, opened 2020-02-04
Add root use of KMS

ahayden:
This isn't a solution, but could be good to have root user access to KMS as an anti-lockout rule.

### DIFF
--- a/templates/essentials.yaml
+++ b/templates/essentials.yaml
@@ -340,6 +340,11 @@ Resources:
             Effect: "Allow"
             Principal:
               AWS:
+                - !Join
+                  - ''
+                  - - 'arn:aws:iam::'
+                    - !Ref AWS::AccountId
+                    - ':root'
                 - !ImportValue us-east-1-bootstrap-TravisUserArn
                 - !ImportValue us-east-1-bootstrap-CfServiceRoleArn
                 - !ImportValue us-east-1-bootstrap-SsmParamLambdaExecutionRoleArn


templates/essentials.yaml
```diff
@@ -340,6 +340,11 @@ Resources:
             Effect: "Allow"
             Principal:
               AWS:
+                - !Join
```
zaro0508: why not just give admin role access?  https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/bootstrap.yaml#L160

if you still prefer root it's better to use a !Sub than a !Join like so..
```
  - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
```

ahayden: Does the admin role exist when this template is built?

zaro0508: yes, bootstrap.yaml is the first template deployed to each account.

ahayden: but the admin role is output from accounts.yaml and so cannot be referenced by bootstrap.yaml... right?

zaro0508: gah, you are correct!! there is no way to give admin role access to this key.  I guess this is ok 
 but root doesn't help because logging in with root requires MFA and i believe the MFA is kept under lock and key.   I see two options:
1. Create a new key in account.yaml and re-encrypt with secure string with the new key
2. All IT admins get the travis creds to update the secrets.

ahayden: I'll merge this, then. I think (1) is the best choice though I need to make sure that string isn't actually used in bootstrap.

PR #226, opened 2020-02-24
Managed rds postgres

ahayden:
This adds RDS postgres option to the provisioner. Every provisioner request will need admin set up of the database credentials.
Passes pre-commit.

### DIFF
--- a/templates/managed-rds-postgresql-v1.j2
+++ b/templates/managed-rds-postgresql-v1.j2
@@ -2,6 +2,25 @@ AWSTemplateFormatVersion: '2010-09-09'
 Description: >-
   Provision RDS PostgreSQL instance
 Parameters:
+  DomainName:
+    Description: Domain name for your website (example.org)
+    Type: String
+    AllowedPattern: (?!-)[a-zA-Z0-9-.]{1,63}(?<!-)
+    ConstraintDescription: must be a valid DNS zone name.
+  SubDomainName:
+    Description: The sub domain name e.g. ('my' in my.example.org)
+    Type: String
+    AllowedPattern: (?!-)[a-zA-Z0-9-.]{1,63}(?<!-)
+    ConstraintDescription: must be a valid DNS zone name.
+  AcmCertificateArn:
+    Type: String
+    Description: The Amazon Resource Name (ARN) of an AWS Certificate Manager (ACM) certificate.
+    AllowedPattern: "arn:aws:acm:.*"
+    ConstraintDescription: must be a valid certificate ARN.
+  Department:
+    Description: 'The department for this resource'
+    Type: String
+    Default: ""
   VpcSubnet:
     Type: List<AWS::EC2::Subnet::Id>
     ConstraintDescription: must be an existing subnet
@@ -16,7 +35,7 @@ Parameters:
   KmsKeyId:
     Description: The Amazon Resource Name (ARN) of the AWS Key Management Service master key that is used to create the encrypted volume
     Type: String
-    Default: ""
+    Default: !Ref AWSKmsInfraKey
   DBPortNumber:
     Description: The port number on which the database accepts connections.
     Type: Number
@@ -103,16 +122,17 @@ Parameters:
       - 'false'
     ConstraintDescription: must be true or false.
 Conditions:
-  HasMksKeyId: !Not [!Equals ["", !Ref KmsKeyId]]
+  HasKmsKeyId: !Not [!Equals ["", !Ref KmsKeyId]]
 Resources:
   DBSubnetGroup:
     Type: AWS::RDS::DBSubnetGroup
     Properties:
       DBSubnetGroupDescription: Subnet group name within accounts
+      # Currently not able to reference private subnets
       SubnetIds:
         - PublicSubnet
-    - PublicSubnet1
-    - PublicSubnet2
+        - PublicSubnet1
+        - PublicSubnet2
       Tags:
         - Key: "Department"
           Value: !Ref Department
--- a/templates/managed-rds-postgresql-v1.j2
+++ b/templates/managed-rds-postgresql-v1.j2
@@ -108,9 +108,9 @@ Resources:
   DBSubnetGroup:
     Type: AWS::RDS::DBSubnetGroup
     Properties:
-      DBSubnetGroupName: !Ref ##TODO VPC-Subnet override
+      DBSubnetGroupName: !Ref 'VpcName'
       DBSubnetGroupDescription: Subnet group name within accounts
-      SubnetIds: !Ref 'VPCSubnet'
+      SubnetIds: !Ref 'VpcSubnet'
   {% if sceptre_user_data.IngressWhitelist is defined %}
   DBInstanceSecurityGroup:
     Type: 'AWS::EC2::SecurityGroup'
@@ -180,10 +180,6 @@ Outputs:
     Value: !Ref DBInstance
     Export:
       Name: !Sub '${AWS::Region}-${AWS::StackName}-DBInstance'
-  OwnerEmail:
-    Value: !Ref OwnerEmail
-    Export:
-      Name: !Sub '${AWS::Region}-${AWS::StackName}-OwnerEmail'
   DBInstanceJDBCConnectionString: #TODO this should potential be the FQDN of the endpoint
     Description: JDBC connection string for the master database
     Value: !Join
--- a/templates/managed-rds-postgresql-v1.j2
+++ b/templates/managed-rds-postgresql-v1.j2
@@ -1,17 +1,13 @@
-AWSTemplateFormatVersion: 2010-09-09
+AWSTemplateFormatVersion: '2010-09-09'
 Description: >-
   Provision RDS PostgreSQL instance
 Parameters:
-  InstanceType:
-    Description: Database instance type
-    Type: String
-    Default: db.t2.small
-  VpcSubnet:
-    Description: Name of an existing VPC subnet to run the instance in.
-    Type: String
+  DBSubnets:
+    Type: List<AWS::EC2::Subnet::Id>
+    Description: Select at least 2 subnets across 2 AZs to include in the database
+      subnet group
+    ConstraintDescription: must be an existing subnet
     Default: PrivateSubnet
-    ConstraintDescription: >-
-      Allowed values (PrivateSubnet, PrivateSubnet1, PrivateSubnet2, PublicSubnet, PublicSubnet1, PublicSubnet2)
     AllowedValues:
       - PrivateSubnet
       - PrivateSubnet1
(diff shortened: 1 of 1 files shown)


templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,284 @@
...
+Resources:
+  {% if sceptre_user_data.IngressWhitelist is defined %}
+  DBInstanceSecurityGroup:
+    Type: 'AWS::EC2::SecurityGroup'
+    Properties:
+      GroupDescription: "Open database on defined port for access by whitelisted IP"
+      VpcId: !ImportValue
+        'Fn::Sub': '${AWS::Region}-${VpcName}-VPCId'
+      SecurityGroupIngress:
+      {% for block in sceptre_user_data.IngressWhitelist %}
```
zaro0508: [optional] would be more clear if `block` is `ip` or `CidrIp`

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,284 @@
...
+    DeletionPolicy: Snapshot
+    Tags:
+      - Key: "Name"
+        Value: !Ref 'AWS::StackName'
+      - Key: "Department"
+        Value: !Ref Department
+      - Key: "Project"
+        Value: Project
+      - Key: "OwnerEmail"
+        Value: !Ref OwnerEmail
```
zaro0508: when you setup instances with an `init` script you need to define a `CreationPolicy`.  The creation policy tells AWS to wait until the script completes.  https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/managed-ec2-linux-v1.j2#L287

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,284 @@
...
+    Condition: PublicDBResource
+    Value: !GetAtt 'DBInstance.Endpoint.Address'
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-DBInstancePublicIp'
+  DBInstancePrivateIp:
+    Condition: PrivateDBResource
+    Value: !GetAtt 'DBInstance.Endpoint.Address'
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-DBInstancePrivateIp'
+  OwnerEmail:
```
zaro0508: You probably don't need this.  it's only used to by the sceptre notification hook.   i think you are going to notify those dudes yourself.

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,284 @@
...
+          - "--tags Key=Name,Value=$DB_INSTANCE_ID-root Key=cloudformation:stack-name,Value=$STACK_NAME "
+          - "Key=Department,Value=$DEPARTMENT Key=Project,Value=$PROJECT Key=OwnerEmail,Value=$OWNER_EMAIL"
+          env:
+            AWS_REGION: !Ref AWS::Region
+            STACK_NAME: !Ref AWS::StackName
+            DEPARTMENT: !Ref Department
+            PROJECT: !Ref Project
+            OWNER_EMAIL: !Ref OwnerEmail
+    Properties:
+      AllocatedStorage: !Ref 'VolumeSize'
```
zaro0508: Since we plan to put this db in the public then would it make sense to setup SSL connection?  https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_PostgreSQL.html#PostgreSQL.Concepts.General.SSL

i think that's what the ` CACertificateIdentifier` is for?  https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-cacertificateidentifier

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,284 @@
...
+      DBInstanceClass: !Ref 'InstanceType'
+      KmsKeyId: !If [HasKmsKeyId, !Ref KmsKeyId, !Ref 'AWS::NoValue']
+      StorageEncryption: !Ref 'EncryptVolume'
+      Engine: postgres
+      EngineVersion: !Ref 'DBEngineVersion'
+      MasterUsername: !Ref 'DBMasterUsername'
+      MasterUserPassword: !Ref 'DBMasterUserPassword'
+      MultiAZ: !Ref 'MultiAZ'
+      BackupRetentionPeriod: !Ref 'DBBackupRetentionPeriod'
```
zaro0508: there's no relationship between this DB and the `DBInstanceSecurityGroup`?  don't we need a `DBSecurityGroups` or `VPCSecurityGroups` property?

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-dbsecuritygroups

zaro0508: I think you need a `DBSubnetGroupName` otherwise it will get put in the default VPC.  You want to put it in the `sandcastlevpc` for org-sagebase-sandbox and `snowflakevpc` in org-sagebase-scicomp account.

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,284 @@
...
+            AWS_REGION: !Ref AWS::Region
+            STACK_NAME: !Ref AWS::StackName
+            DEPARTMENT: !Ref Department
+            PROJECT: !Ref Project
+            OWNER_EMAIL: !Ref OwnerEmail
+    Properties:
+      AllocatedStorage: !Ref 'VolumeSize'
+      DBInstanceClass: !Ref 'InstanceType'
+      KmsKeyId: !If [HasKmsKeyId, !Ref KmsKeyId, !Ref 'AWS::NoValue']
+      StorageEncryption: !Ref 'EncryptVolume'
```
zaro0508: this should be `StorageEncrypted`

zaro0508: `StorageType` seems important, might want to see if those dudes care about that.

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,284 @@
...
+      - '10.6'
+      - '10.5'
+      - '10.4'
+      - '10.3'
+      - '10.1'
+  InstanceType:
+    Description: The database instance type
+    Type: String
+    Default: db.t2.small
+    AllowedValues:
```
zaro0508: are all these instance types really possible?  https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/Concepts.DBInstanceClass.html#Concepts.DBInstanceClass.Types

ahayden: yah, they should all be possible but I paired the list back to just a few reasonable ones.

templates/managed-rds-postgresql-v1.j2
ahayden: yah could be... I'll leave it as standard for now and we can revisit it if they have a need for optimization.

templates/managed-rds-postgresql-v1.j2
ahayden: yes... there is a `VPCSecurityGroups` property but is has a typo.

templates/managed-rds-postgresql-v1.j2
ahayden: I'm hoping Andrew or James can request it through the provisioner

ahayden: But, I'm removing those as the JDBC Connection string is sufficient.

templates/managed-rds-postgresql-v1.j2
ahayden: IO'll use CIDRBlock to be specific

templates/managed-rds-postgresql-v1.j2
ahayden: Nice- I believe I took the `DeletionPolicy` clause form the a SC template

templates/managed-rds-postgresql-v1.j2
ahayden: I forgot about this step. I'll add the `CACertificateIdentifier` field but it'll probably be a manual step

templates/managed-rds-postgresql-v1.j2
ahayden: What if the subnet name is in the highest level template? Like in the provisioner request: https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/303

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,252 @@
...
+            STACK_NAME: !Ref AWS::StackName
+            DEPARTMENT: !Ref Department
+            PROJECT: !Ref Project
+            OWNER_EMAIL: !Ref OwnerEmail
+    Properties:
+      DBInstanceClass: !Ref 'InstanceType'
+      AllocatedStorage: !Ref 'VolumeSize'
+      StorageType: !Ref 'StorageType'
+      StorageEncrypted: !Ref 'EncryptVolume'
+      CACertificateIdentifier: !ImportValue: 'Fn::Sub': '${AWS::Region}-${AWS::StackName}-CertificateArn'
```
zaro0508: this doesn't seem valid?  where is `${AWS::Region}-${AWS::StackName}-CertificateArn` ?

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,252 @@
...
+          env:
+            AWS_REGION: !Ref AWS::Region
+            STACK_NAME: !Ref AWS::StackName
+            DEPARTMENT: !Ref Department
+            PROJECT: !Ref Project
+            OWNER_EMAIL: !Ref OwnerEmail
+    Properties:
+      DBInstanceClass: !Ref 'InstanceType'
+      AllocatedStorage: !Ref 'VolumeSize'
+      StorageType: !Ref 'StorageType'
```
zaro0508: if you change this later it will require an interruption, this may mean downtime or migration.  you should really figure this out at the beginning.

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,252 @@
...
+        owner: root
+        group: root
+    commands:
+      01_enable_cfn-hup:
+        command: "/bin/systemctl enable cfn-hup.service"
+      02_start_cfn-hup:
+        command: "/bin/systemctl start cfn-hup.service"
+        tag_volume:
+      commands:
+        01_tag_root_disk:
```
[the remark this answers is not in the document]
zaro0508: We tag root volume in the provisioner because we allow users to set the root volume size so we want to track the cost.  When provisioning an RDS the root volume is most likely very small and most likely ephermeral so there is no additional cost.  My suggestion is to just remove this entire `AWS::CloudFormation::Init` and don't worry about tagging the root volume.

templates/managed-rds-postgresql-v1.j2
ahayden: Removing... I'll resolve this comment.

templates/managed-rds-postgresql-v1.j2
ahayden: Makes sense to me. Can that field take wildcards? I limited it further than that, allowing only larger versions of the optimized instance types since there's really no reason to have a small, memory-optimized instance for our cases.

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,252 @@
...
+      CACertificateIdentifier: !ImportValue: 'Fn::Sub': '${AWS::Region}-${AWS::StackName}-CertificateArn'
+      KmsKeyId: !If [HasKmsKeyId, !Ref KmsKeyId, !Ref 'AWS::NoValue']
+      Engine: postgres
+      EngineVersion: !Ref 'DBEngineVersion'
+      MasterUsername: !Ref 'DBMasterUsername'
+      MasterUserPassword: !Ref 'DBMasterUserPassword'
+      MultiAZ: !Ref 'MultiAZ'
+      BackupRetentionPeriod: !Ref 'DBBackupRetentionPeriod'
+      VPCSecurityGroups:
+        - !GetAtt 'DBInstanceSecurityGroup.GroupId'
```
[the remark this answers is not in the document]
ahayden: OK- I will make that field required in the higher-level templates.

templates/managed-rds-postgresql-v1.j2
ahayden: I see your point... Maybe I will make a note in the higher-level template about destruction/migration-requiring changes.

templates/managed-rds-postgresql-v1.j2
ahayden: I missed adding this as a parameter

templates/managed-rds-postgresql-v1.j2
zaro0508: No, you don't need to do that because it's already in the AWS docs.  What I'm suggesting is that you want to make sure this is decided up front when provisioning the DB otherwise it will be a headache to change.

templates/managed-rds-postgresql-v1.j2
zaro0508: I still don't see it?

templates/managed-rds-postgresql-v1.j2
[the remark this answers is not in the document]
zaro0508: can you make a sceptre parameter a `required` parameter?

templates/managed-rds-postgresql-v1.j2
zaro0508: huh? it's not removed?

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,197 @@
...
+      - 'true'
+      - 'false'
+    ConstraintDescription: must be true or false.
+Conditions:
+  HasMksKeyId: !Not [!Equals ["", !Ref KmsKeyId]]
+Resources:
+  DBSubnetGroup:
+    Type: AWS::RDS::DBSubnetGroup
+    Properties:
+      DBSubnetGroupName: !Ref ##TODO VPC-Subnet override
```
zaro0508: can't have this, it needs to be set to something

templates/managed-rds-postgresql-v1.j2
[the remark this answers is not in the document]
ahayden: maybe not "required" for the build to succeed, but required in the sense that it won't be approved without it

templates/managed-rds-postgresql-v1.j2
ahayden: Right- the requestor should set this value for their project specifically

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,185 @@
...
+    ConstraintDescription: must be true or false.
+Conditions:
+  HasMksKeyId: !Not [!Equals ["", !Ref KmsKeyId]]
+Resources:
+  DBSubnetGroup:
+    Type: AWS::RDS::DBSubnetGroup
+    Properties:
+      DBSubnetGroupName: !Ref 'VpcName'
+      DBSubnetGroupDescription: Subnet group name within accounts
+      SubnetIds: !Ref 'VpcSubnet'
```
zaro0508: SubnetIds should be a list of strings not a single string.  Typically you want to provide multiple subnets so it can be redundant.  For this i would set it to...

```
SubnetIds:
  - PublicSubnet
  - PublicSubnet1
  - PublicSubnet2
```

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,185 @@
...
+    AllowedValues:
+      - 'true'
+      - 'false'
+    ConstraintDescription: must be true or false.
+Conditions:
+  HasMksKeyId: !Not [!Equals ["", !Ref KmsKeyId]]
+Resources:
+  DBSubnetGroup:
+    Type: AWS::RDS::DBSubnetGroup
+    Properties:
```
zaro0508: please add tags property and add Department, Project, and OwnerEmail tags to this

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,185 @@
...
+          Value: !Ref OwnerEmail
+  {% endif %}
+  DBInstance:
+    Type: 'AWS::RDS::DBInstance'
+    Properties:
+      DBInstanceClass: !Ref 'InstanceType'
+      AllocatedStorage: !Ref 'VolumeSize'
+      StorageType: !Ref 'StorageType'
+      StorageEncrypted: !Ref 'EncryptVolume'
+      CACertificateIdentifier: !Ref 'AcmCertificateArn'
```
zaro0508: what is this referencing?  i don't see the source?

templates/managed-rds-postgresql-v1.j2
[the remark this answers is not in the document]
zaro0508: Well i mean you can enforce it with jinja. just put a condition around this? like above,  `{% if sceptre_user_data.IngressWhitelist is defined %}`

templates/managed-rds-postgresql-v1.j2
```diff
@@ -0,0 +1,185 @@
...
+        - Key: "Department"
+          Value: !Ref Department
+        - Key: "Project"
+          Value: !Ref Project
+        - Key: "OwnerEmail"
+          Value: !Ref OwnerEmail
+  {% endif %}
+  DBInstance:
+    Type: 'AWS::RDS::DBInstance'
+    Properties:
```
zaro0508: What happened to my suggestion of adding CopyTagsToSnapshot property?  I could have sworn i made this suggestion in one of my earlier reviews... https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-rds-database-instance.html#cfn-rds-dbinstance-copytagstosnapshot

ahayden: I don't remember that in prior reviews, but it makes sense.

templates/managed-rds-postgresql-v1.j2
ahayden: a reference from the provisioner repo: https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/303

PR #227, opened 2020-03-07
Quarantine sg

ahayden:
Adds security group options


templates/vpc.yaml
```diff
@@ -502,6 +502,31 @@ Resources:
           Value: !Ref Project
         - Key: "OwnerEmail"
           Value: !Ref OwnerEmail
+  QuarantineSecurityGroup:
+    Type: 'AWS::EC2::SecurityGroup'
+    Properties:
+      GroupDescription: Security Group for bastians
```
zaro0508: is this accurate, shouldn't it be about quarentining?

templates/vpc.yaml
```diff
@@ -502,6 +502,31 @@ Resources:
...
+      VpcId:
+        Ref: "VPC"
+      SecurityGroupIngress:
+        - CidrIp: "10.0.0.0/8"
```
zaro0508: shouldn't this be ingress from our VPN?

templates/vpc.yaml
ahayden: yeah, updated

templates/vpc.yaml
ahayden: it's shorter this way... but i updated it with the VPN range instead

PR #233, opened 2020-04-01
Add scheduled maintenance tasks and add managed policy to instance profile

ahayden:
The objective is to create a maintenance routine that operates on a cron schedule to run two tasks against EC2s with specific tags on them. The first task is to update the SSM agent on the instance;
the second is to run the AWS default patch baseline task.

This would schedule only one maintenance window: 10 am UTC every other Sunday, to run for 3 hours and reboot if needed

Depends:

Adding SSM policy to instance profiles of instances generated through SC: Sage-Bionetworks/scipoolprod-sc-lib-infra#106
Add instance profile option for instances not created in SC (not related to SC-12):
Sage-Bionetworks/aws-infra#233

### DIFF
--- a/templates/essentials.yaml
+++ b/templates/essentials.yaml
@@ -179,38 +179,6 @@ Resources:
           Value: !Ref Project
         - Key: "OwnerEmail"
           Value: !Ref OwnerEmail
-  AWSS3SSMTasklogBucket:
-    Type: AWS::S3::Bucket
-    Properties:
-      Tags:
-        - Key: "Department"
-          Value: !Ref Department
-        - Key: "Project"
-          Value: !Ref Project
-        - Key: "OwnerEmail"
-          Value: !Ref OwnerEmail
-  AWSSSMTaskNotificationSNSTopic:
-    Type: "AWS::SNS::Topic"
-    Properties:
-      Subscription:
-        -
-          Endpoint: !Ref OperatorEmail
-          Protocol: email
-  AWSSSMTaskNotificationSNSTopicPolicy:
-    Type: "AWS::SNS::TopicPolicy"
-    Properties:
-      Topics:
-        - !Ref AWSSSMTaskNotificationSNSTopic
-      PolicyDocument:
-        Version: "2012-10-17"
-        Statement:
-          -
-            Sid: "SSMTaskNotificaitonSNSTopicPolicy"
-            Effect: "Allow"
-            Principal:
-              Service: "ssm.amazonaws.com"
-            Resource: "*"
-            Action: "SNS:Publish"
   # Bucket for lambda artifacts
   AWSS3LambdaArtifactsBucket:
     Type: AWS::S3::Bucket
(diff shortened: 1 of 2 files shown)


templates/essentials.yaml
```diff
@@ -387,11 +388,16 @@ Resources:
...
+                #Needed for SSM management
+                - "ssm.amazonaws.com"
             Action:
               - "sts:AssumeRole"
       Path: "/"
       ManagedPolicyArns:
         - !Ref TagRootVolumePolicy
+          #Add policy to role allow SSM to attach to instances
+        - 'arn:aws:iam::aws:policy/AmazonSSMManagedInstanceCore'
+# This profile would more accurately be named "EC2ManagementProfile"
```
zaro0508, line 400: I believe this change will re-provision existing instances.

How about we create a new set of `EC2ManagementProfile`, `EC2ManagementRole` and `EC2ManagementPolicy`.  Then selectively apply this to specific templates?

ahayden: We won't need a `EC2ManagementPolicy` for this step as I'm using an AWS managed policy, but I agree, this should be separated into a new role to prevent a potentially destructive change

templates/essentials.yaml
```diff
@@ -466,6 +492,16 @@ Outputs:
...
+  ManagedInstanceProfile:
+    Description: Profile to allow instances to tag its root volume and allow SSM actions
+    Value: !Ref ManagedInstanceProfile
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ManagedInstanceProfile'
+  ManagedInstanceRole:
+    Description: Role to allow instances to tag its root volume
+    Value: !Ref ManagedInstanceRole
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ManagedInstanceRole'
```
zaro0508, line 504: These are outputs of the logical names of the resource however ARNs typically useful than names.  I would suggest either replacing these outputs with the ARN or adding outputs for ARNs.  for example..

```
  ManagedInstanceRoleArn:
    Description: Role ARN to allow instances to tag its root volume
    Value: !GetAtt ManagedInstanceRole.Arn
    Export:
      Name: !Sub '${AWS::Region}-${AWS::StackName}-ManagedInstanceRoleArn'
```

ahayden: OK, I've add ARN strings as exports for some resources. I wonder if we would hit a limit doubling up on exports.

templates/essentials.yaml
```diff
@@ -466,6 +492,26 @@ Outputs:
...
+    Value: !GetAtt ManagedInstanceProfile.Arn
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ManagedInstanceProfileArn'
+  ManagedInstanceRole:
+    Description: Role to allow instances to tag its root volume
+    Value: !Ref ManagedInstanceRole
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ManagedInstanceRole'
+  ManagedInstanceRoleArn:
+    Description: Role to allow instances to tag its root volume
```
zaro0508, line 511: here too.

templates/essentials.yaml
```diff
@@ -466,6 +492,26 @@ Outputs:
...
+    Value: !Ref ManagedInstanceProfile
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ManagedInstanceProfile'
+  ManagedInstanceProfileArn:
+    Description: Profile to allow instances to tag its root volume and allow SSM actions
+    Value: !GetAtt ManagedInstanceProfile.Arn
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-ManagedInstanceProfileArn'
+  ManagedInstanceRole:
+    Description: Role to allow instances to tag its root volume
```
zaro0508, line 506: missing the SSM part?

ahayden: Agreed and updated

PR #234, opened 2020-04-01
Generic public VPC template

zaro0508:
Add a generic template to create a public VPC with
6 availability zones in 6 subnets.

### DIFF
--- a/templates/managed-public-vpc.yaml
+++ b/templates/managed-public-vpc.yaml
@@ -1,0 +1,340 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >-
+  A public VPC with 6 availability zones and 6 subnets
+Parameters:
+  VpcSubnetPrefix:
+    Description: The VPC subnet prefix (i.e. 10.40)
+    Type: String
+  AZRegions:
+    Description: Availability zone regions
+    Type: List<AWS::EC2::AvailabilityZone::Name>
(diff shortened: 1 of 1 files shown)


templates/managed-public-vpc.yaml
```diff
@@ -0,0 +1,344 @@
...
+  OutboundPublicNetworkAclEntry:
+    Type: "AWS::EC2::NetworkAclEntry"
+    Properties:
+      NetworkAclId:
+        Ref: "PublicNetworkAcl"
+      RuleNumber: 100
+      Protocol: -1
+      RuleAction: "allow"
+      Egress: true
+      CidrBlock: "0.0.0.0/0"
```
ahayden, line 235: Should this ACL deny ingress to private subnets? If the goal is the isolate the hosts in these VPCs

zaro0508: I'm not sure I understand your question.  There aren't any private subnets defined in this VPC and this specific Network ACL is for outbound traffic not inbound traffic.  Also this VPC is generic, similar to the AWS default VPC, it can be used for many different purposes it just depends on how we use it.  This PR is not using the VPC for any specific purpose, it's just defining one.

zaro0508:
> So it's just 6 public zones, like we had before VPCs?

yes, that is correct @xschildw

PR #236, opened 2020-04-03
Add cloudformation macro for setting the OwnerEmail tag on a bucket

tthyer:
This sets up a macro that pulls current tags, derives the email from the Synapse id, adds that tag, reapplies tags.
Uses an aws call that requires that we set all the tags, not just adding the one. 
Only sets the one tag atm but anticipate setting others, too.


lambdas/cfn-set-bucket-tags/README.md
```diff
@@ -0,0 +1,89 @@
...
+applications. It uses Docker to run your functions in an Amazon Linux
+environment that matches Lambda. It can also emulate your application's build
+environment and API.
+
+## Use the SAM CLI to build and test locally
+
+Build your application with the `sam build --use-container` command.
+
+```bash
+cfn-set-bucket-tags$ sam build --use-container
```
zaro0508, line 35: is `cfn-set-bucket-tags$` part of the command to build?

lambdas/cfn-set-bucket-tags/README.md
```diff
@@ -0,0 +1,89 @@
...
+```
+
+The SAM CLI installs dependencies defined in `set_bucket_tags/requirements.txt`,
+creates a deployment package, and saves it in the `.aws-sam/build` folder.
```
zaro0508, line 39: is this really needed? can't we expect lambda developers to know this?

.travis.yml
```diff
@@ -7,6 +7,7 @@ install:
   - unzip -j -n /tmp/infra-utils.zip -x "infra-utils-master/.gitignore" "infra-utils-master/LICENSE" "infra-utils-master/*.md" "infra-utils-master/aws/*"
   - ./setup_aws_cli.sh || travis_terminate 1
   - pip install pre-commit aws-sam-cli
+  - pip install -r lambdas/cfn-set-bucket-tags/set_bucket_tags/requirements.txt
```
zaro0508, line 10: is this so you can run your tests?  it's inappropriate to install that here.

.travis.yml
```diff
@@ -20,41 +21,17 @@ jobs:
       script:
         - pre-commit autoupdate
         - pre-commit run --all-files
-        - sam build --build-dir lambdas/build/jumpcloud --base-dir lambdas/jumpcloud --template lambdas/jumpcloud/template.yaml
-        - sam build --build-dir lambdas/build/rotate-credentials --base-dir lambdas/rotate-credentials --template lambdas/rotate-credentials/template.yaml
-        - sam build --build-dir lambdas/build/cfn-explode-macro --base-dir lambdas/cfn-explode-macro --template lambdas/cfn-explode-macro/template.yaml
-        - sam build --build-dir lambdas/build/cfn-ssm-param-macro --base-dir lambdas/cfn-ssm-param-macro --template lambdas/cfn-ssm-param-macro/template.yaml
-        - sam build --build-dir lambdas/build/cfn-s3objects-macro --base-dir lambdas/cfn-s3objects-macro --template lambdas/cfn-s3objects-macro/template.yaml
-        - sam build --build-dir lambdas/build/cfn-oidc-identity-provider --base-dir lambdas/cfn-oidc-identity-provider --template lambdas/cfn-oidc-identity-provider/template.yaml
+        - ./build-lambdas.sh
```
zaro0508, line 24: if your trying to change the travis build then i would prefer this to happen in a separate PR

lambdas/cfn-set-bucket-tags/README.md
```diff
@@ -0,0 +1,89 @@
...
+
+## Use in a Cloudformation Template
+Insert in your Cloudformation template a block similar to this as a sibling
+of the S3 bucket (called in this example `S3Bucket`) :
+```yaml
+  'Fn::Transform':
+    Name: "SetBucketTags"
+    Parameters:
+      BucketName: !Ref S3Bucket
+```
```
zaro0508, line 20: It would help to have a more complete example so we can see the relationship between S3 and this transform.  I would suggest to have an example that includes the S3Bucket definition, transform and the expected transformed template that will be produced.

.travis.yml
tthyer: In that case we need to discuss how we are going to continue to develop separate python applications with possibly conflicting requirements in the same repo and run their tests.

.travis.yml
tthyer: Ok, I'll move this.

lambdas/cfn-set-bucket-tags/README.md
tthyer: The template is not transformed. I will add some explanatory text to make this clear.

lambdas/cfn-set-bucket-tags/README.md
tthyer: No, this is the prompt, and it's in the README you get from running `sam init`.

lambdas/cfn-set-bucket-tags/README.md
tthyer: Again, some of this background text comes from `sam init`.

tthyer:
Closing as this code is being moved to https://github.com/Sage-Bionetworks/it-lambda-set-bucket-tags

PR #239, opened 2020-04-15
[SYNPY-1054] Allow all S3 VPC Endpoints to access same region restricted S3 Buckets

jkiang13:
https://sagebionetworks.jira.com/browse/SYNPY-1054

A user in a different VPC in us-east-1 with a configured S3 VPC Endpoint Gateway is being denied access to the amp-mayo-sinai-synapseencryptedexternalbucket-1bmvn8rlwixv2 S3 bucket.

This is a modification of the behavior for [IT-661](https://github.com/Sage-Bionetworks/aws-infra/pull/173) which allowed access specifically via the S3 VPC Endpoint for the Sage computevpc.

S3 VPC Endpoints only support intra-region traffic and any S3 traffic with an aws:sourceVpc is implicitly from the same region. This changes the behavior of the region block to allow any VPC endpoint sourced request, not just a list of explicitly defined VPCs. Originally I had thought to further parameterize SameRegionResourceAccessToBucket, but I can't think of a case where we'd want to allow requests from all a region's public IPs but only allow a subset of private VPC traffic from that region so this removes the existing parameterization in favor of a general VPC endpoint permission.

https://docs.aws.amazon.com/vpc/latest/userguide/vpce-gateway.html
> Endpoints are supported within the same Region only. You cannot create an endpoint between a VPC and a service in a different Region.
> other internet traffic goes to your internet gateway, including traffic that's destined for other services, **and destined for the AWS service in other Regions**

### DIFF
--- a/templates/AddSameRegionBucketDownloadRestriction.yaml
+++ b/templates/AddSameRegionBucketDownloadRestriction.yaml
@@ -27,12 +27,6 @@ Parameters:
     AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
     ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
     Default: 'it@sagebase.org'
-  # This lambda, by default, restricts access to buckets when Gateway VPC Endpoints are enabled on an AWS account
-  # This parameter overrides that by explicitly granting bucket access to resources in this list of VPCs.
-  SameRegionResourceAccessToBucketGrantVpcAccess:
-    Type: CommaDelimitedList
-    Description: Allow resources in a VPC (i.e. ["vpc-111bbb22", "vpc-123add45"]) to access the bucket
-    ConstraintDescription: Comma separated list of VPC IDs (i.e. ["vpc-111bbb22", "vpc-123add45"])
 Resources:
   ExternalBucketGroupPolicyUpdateRole:
     Type: "AWS::IAM::Role"
@@ -76,7 +70,6 @@ Resources:
       Environment:
         Variables:
           BUCKET_NAME: !Ref "BucketName"
-          VPC_IDS: !Join [ ",", !Ref "SameRegionResourceAccessToBucketGrantVpcAccess"]
           REGION: !Ref "AWS::Region"
       Code:
         ZipFile: |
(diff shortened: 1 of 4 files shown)


templates/AddSameRegionBucketDownloadRestriction.yaml
```diff
@@ -27,12 +27,6 @@ Parameters:
     AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
     ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
     Default: 'it@sagebase.org'
-  # This lambda, by default, restricts access to buckets when Gateway VPC Endpoints are enabled on an AWS account
-  # This parameter overrides that by explicitly granting bucket access to resources in this list of VPCs.
-  SameRegionResourceAccessToBucketGrantVpcAccess:
```
zaro0508, line 32: there are tests that reference this, you'll need to remove from here: https://github.com/Sage-Bionetworks/aws-infra/blob/master/ci/AddSameRegionBucketDownloadRestriction-input.json#L7

Need to remove dependency here: https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/config/prod/AMP-Mayo-Sinai-Bucket.yaml#L28

templates/s3-bucket.yaml
```diff
@@ -26,12 +26,6 @@ Parameters:
       - true
       - false
     Default: false
-  # only used with SameRegionResourceAccessToBucket, more info in AddSameRegionBucketDownloadRestriction.yaml
-  SameRegionResourceAccessToBucketGrantVpcAccess:
```
zaro0508, line 30: since it's removed well need to update our PR examples that mention this parameter:

https://github.com/Sage-Bionetworks/sandbox-provisioner/pull/4/files
https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/14/files

templates/AddSameRegionBucketDownloadRestriction.yaml
jkiang13: Removed the stale test param. Is there a way to see/run the taskcat tests?

The dependency is removed in the linked PR: https://github.com/Sage-Bionetworks/scicomp-provisioner/pull/317

templates/s3-bucket.yaml
jkiang13: I don't see either the now removed SameRegionResourceAccessToBucketGrantVpcAccess, or the remaining SameRegionResourceAccessToBucket parameter in those PR examples. Would you like me to create a PR to your forked iwanna-bucket branches to add the latter?

zaro0508: sorry, you are correct.  I forgot we never added info about this since it's a rarely requested.

templates/AddSameRegionBucketDownloadRestriction.yaml
zaro0508: unfortunately taskcat only runs on merge because it needs AWS secrets to run.

jkiang13:
> is the description in AddSameRegionBucketDownloadRestriction.yaml still accurate. Do you think it needs updating for this change?

Yes it's still accurate. It will now serve to accept private intra region via VPC endpoint requests, but the fact that it wouldn't before wasn't implied by the description, so it's now more true to the existing description.

PR #241, opened 2020-04-28
Allow calling Get-Role to ManagedInstanceProfile

ahayden:
Adding this policy to the ManagedInstanceProfile allows instances to query for the roleid of assumed role users so they can add the roleid to a tag on themselves. This is needed for the SSM StartSession connection filter that restricts access to specific roleid/sessionname strings. I've changed this PR to allow reading role on any string, "*", because the rolename could have any pattern.

Merging this PR might destructively change instances with the ManagedInstance Profile, which should be limited to Service Catalog instances.  
Passes pre-commit

### DIFF
--- a/templates/essentials.yaml
+++ b/templates/essentials.yaml
@@ -400,7 +400,7 @@ Resources:
             Effect: Allow
             Action:
               - "iam:GetRole"
-            Resource: !Sub 'arn:aws:sts:${AWS::Region}:${AWS::AccountId}:assumed-role/*'
+            Resource: "*"
   # Allow instances to apply tags to its root volume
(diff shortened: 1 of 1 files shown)


tthyer:
Note, according to the docs changing the IamInstanceProfile value on an instance won't do a replace:
```IamInstanceProfile
The IAM instance profile. See IamInstanceProfileSpecification in the Amazon EC2 API Reference for property values.

Required: No

Type: String

Update requires: No interruption```

https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html

ahayden:
Closing as we only need this permission in SC

PR #244, opened 2020-05-02
SC-12, moved to general aws-infra

ahayden:
This defines a single scheduled task that runs in an account every week and operates on specifically tagged resources.

### DIFF
--- a/templates/maintenance.yaml
+++ b/templates/maintenance.yaml
@@ -4,11 +4,11 @@ Resources:
   DefaultMaintenanceWindow:
     Type: 'AWS::SSM::MaintenanceWindow'
     Properties:
-      Name: test-ahayden-ProdDefaultMaintenanceWindow
-      Description: Automated deliver patches every other Sunday
+      Name: prod-default-maintenance-window
+      Description: Automated deliver patches every Sunday
       Duration: 3
       Cutoff: 1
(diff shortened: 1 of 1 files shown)


templates/maintenance.yaml
```diff
@@ -4,11 +4,11 @@ Resources:
...
     Type: 'AWS::SSM::MaintenanceWindow'
     Properties:
-      Name: ProdDefaultMaintenanceWindow
-      Description: Automated deliver patches every other Sunday
+      Name: prod-default-maintenance-window
+      Description: Automated deliver patches every Sunday
       Duration: 3
       Cutoff: 1
-      Schedule: 'cron(00 10 ? * SUN/2 *)' #1000UTC Every other Sunday
+      Schedule: 'cron(00 10 ? * SUN *)' #1000UTC every Sunday
```
zaro0508, line 11: why can't this be every other week again?  `SUN/2` expression seems wrong to me.  could you do every 14 days instead, like `00 10 */14 * SUN *`?  or i think you can probably even use the rate schedule, `rate(14 days)`

ahayden: I don't think those expressions evaluate to what an enduser would expect as "every other Sunday".  Doesn't it seem like it would be less predictable for an enduser if they had to figure out which weeks were firing and which were not?

zaro0508: i see, so every 14 days is not predictable.

PR #245, opened 2020-05-13
Add protected tag path rule

ahayden:
This change does not touch the essentials policy that allows tagging, so instances should continue to be able to tag resources associated with them (and other's resources), but it adds a deny rule for tag paths that can be overridden by the instance making the call to create a tag and not by other instances.

Passes pre-commit

### DIFF
--- a/templates/essentials.yaml
+++ b/templates/essentials.yaml
@@ -389,7 +389,7 @@ Resources:
       Path: "/"
       Roles:
         - !Ref ManagedInstanceRole
-  # Allow instances to apply tags to its root volume
+  # Allow instances to apply tags to anything and deny write to TagKeys like "Protected/*" except by self
   TagRootVolumePolicy:
     Type: "AWS::IAM::ManagedPolicy"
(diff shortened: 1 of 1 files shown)


templates/essentials.yaml
```diff
@@ -402,6 +402,17 @@ Resources:
               - "ec2:Describe*"
               - "ec2:CreateTags"
             Resource: "*"
+          - Sid: DenyWriteProtectedPath
```
zaro0508, line 405: changing this has a cascading effect of updating TagRootVolumeProfile and TagRootVolumeRole.  My concern is that this may cause any instances using TagRootVolumeProfile and TagRootVolumeRole directly to re-provision.  Have you tested that?

ahayden: I have tested whether an instance rebuilds after a change to a policy referenced in its instance profile, and it does not force a rebuild.

zaro0508: sound good

PR #250, opened 2020-06-26
Define mini VPC template with small subnets

ahayden:
Here's one option for creating VPC peers for projects that don't use up or overlap address space among peer VPCs. This just takes our regular VPC template and reduces its subnets by one order of CIDR input. I'm suggesting that we reserve the high ranges of our address space for small subnets like this, e.g. 10.255.1..255.0/24 can fit a lot of discrete subnets. Private ranges are /26 and public ranges are /28 based on the assumption that resources are more frequently private.

We would use it like this:
````
template_path: remote/vpc-mini.yaml
stack_name: newvpcnameforproject
dependencies:
  - prod/bootstrap.yaml
parameters:
  VpcSubnetPrefix: "10.255.20"
  VpcName: newvpcnameforproject
hooks:
  before_launch:
    - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/master/vpc-mini.yaml --create-dirs -o templates/remote/vpc-mini.yaml"
````

### DIFF
--- a/templates/vpc-mini.yaml
+++ b/templates/vpc-mini.yaml
@@ -8,7 +8,7 @@ Parameters:
   VpcSubnetPrefix:
     Description: The VPC subnet prefix (i.e. 10.255.40)
     Type: String
-    AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){2}([0-9]|[1-9][0-9]|1[0-9]    {2}|2[0-4][0-9]|25[0-5])$
+    AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){2}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$
   PublicSubnetZones:
     Description: Availability zones for public subnets
     Type: List<AWS::EC2::AvailabilityZone::Name>
--- a/templates/vpc-mini.yaml
+++ b/templates/vpc-mini.yaml
@@ -40,7 +40,7 @@ Mappings:
     Public2:
       CIDR: "160/28" #10.255.X.160-175
     Private2:
-      CIDR: "176/26" #10.255.X.176-240
+      CIDR: "176/26" #10.255.X.176-239
 Resources:
   VPC:
     Type: "AWS::EC2::VPC"
--- a/templates/vpc-mini.yaml
+++ b/templates/vpc-mini.yaml
@@ -8,6 +8,7 @@ Parameters:
   VpcSubnetPrefix:
     Description: The VPC subnet prefix (i.e. 10.255.40)
     Type: String
+    AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){2}([0-9]|[1-9][0-9]|1[0-9]    {2}|2[0-4][0-9]|25[0-5])$
   PublicSubnetZones:
     Description: Availability zones for public subnets
     Type: List<AWS::EC2::AvailabilityZone::Name>
@@ -24,24 +25,6 @@ Parameters:
     Default: "10.1.0.0/16"
     AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(1[6-9]|2[0-8]))$
     ConstraintDescription: CIDR block parameter must be in the form x.x.x.x/16-28
-  Department:
-    Description: 'The department for this resource'
-    Type: String
-    AllowedPattern: '^\S*$'
-    ConstraintDescription: 'Must be string with no spaces'
-    Default: 'Platform'
-  Project:
-    Description: 'The name of the project that this resource is used for'
-    Type: String
-    AllowedPattern: '^\S*$'
-    ConstraintDescription: 'Must be string with no spaces'
-    Default: 'Infrastructure'
-  OwnerEmail:
-    Description: 'Email address of the owner of this resource'
-    Type: String
-    AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
-    ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
-    Default: 'it@sagebase.org'
 Mappings:
   SubnetConfig:
     VPC:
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: templates/vpc-mini.yaml [hunk 1]
```diff
@@ -0,0 +1,628 @@
...
+  OwnerEmail:
+    Description: 'Email address of the owner of this resource'
+    Type: String
+    AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
+    ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
+    Default: 'it@sagebase.org'
+Mappings:
+  SubnetConfig:
+    VPC:
+      CIDR: "0/24" #10.255.X.0/24 = 10.255.X.0-10.255.X.255
```
zaro0508, line 48: Three things..

1. I think CIDR needs to be in `X.X/YY` notation? 
2. CIDR "0.0/24" only allows 164 hosts per subnet.  This seem very small, maybe too small?  But seems ok as well.
3. Unless your intentionally putting it in non-contiguous blocks then then the  `CIDR: "0.0/24"` with the first 6 subnets it should be.. 
```
Network | Start | End | Hosts | Netmask
10.20.0.0/28 | 10.20.0.1 | 10.20.0.14 | 14 | 255.255.255.240
10.20.0.16/28 | 10.20.0.17 | 10.20.0.30 | 14 | 255.255.255.240
10.20.0.32/28 | 10.20.0.33 | 10.20.0.46 | 14 | 255.255.255.240
10.20.0.48/28 | 10.20.0.49 | 10.20.0.62 | 14 | 255.255.255.240
10.20.0.64/28 | 10.20.0.65 | 10.20.0.78 | 14 | 255.255.255.240
10.20.0.80/28 | 10.20.0.81 | 10.20.0.94 | 14 | 255.255.255.240
```

ahayden: IIUC the mapping appends the value of `CIDR` to the input parameter `VpcSubnetPrefix` which is the first three octets of an IPv4 address`10.20.0` in a mini vpc. Our normal VPC template takes the first two octets as an input parameter. In the mini vpc template this results in strings that make up full CIDR blocks of different sizes, ie:
```
VPC contains all 10.20.0.0/24 = 10.20.0.0->10.20.0.255
PublicSubnet is 10.20.0.0/28 = 10.20.0.0->10.20.0.15
PrivateSubnet is 10.20.0.16/26 = 10.20.0.16->10.20.0.79
PublicSubnet1 is 10.20.0.80/28 = 10.20.0.80->10.20.0.95
PrivateSubnet1 is 10.20.0.96/26 = 10.20.0.96->10.20.0.159
PublicSubnet2 is 10.20.0.160/28 = 10.20.0.160->10.20.0.175
PrivateSubnet2 is 10.20.0.176/26 = 10.20.0.176->10.20.0.239
```

zaro0508: Here's the full addressed subnets for CIDR 10.20.0/24 via https://community.spiceworks.com/tools/subnet-calc/
```
Network | Start | End | Hosts | Netmask
-- | -- | -- | -- | --
10.20.0.0/28 | 10.20.0.1 | 10.20.0.14 | 14 | 255.255.255.240Remove
10.20.0.16/28 | 10.20.0.17 | 10.20.0.30 | 14 | 255.255.255.240Remove
10.20.0.32/28 | 10.20.0.33 | 10.20.0.46 | 14 | 255.255.255.240Remove
10.20.0.48/28 | 10.20.0.49 | 10.20.0.62 | 14 | 255.255.255.240Remove
10.20.0.64/28 | 10.20.0.65 | 10.20.0.78 | 14 | 255.255.255.240Remove
10.20.0.80/28 | 10.20.0.81 | 10.20.0.94 | 14 | 255.255.255.240Remove
10.20.0.96/28 | 10.20.0.97 | 10.20.0.110 | 14 | 255.255.255.240Remove
10.20.0.112/28 | 10.20.0.113 | 10.20.0.126 | 14 | 255.255.255.240Remove
10.20.0.128/28 | 10.20.0.129 | 10.20.0.142 | 14 | 255.255.255.240Remove
10.20.0.144/28 | 10.20.0.145 | 10.20.0.158 | 14 | 255.255.255.240Remove
10.20.0.160/28 | 10.20.0.161 | 10.20.0.174 | 14 | 255.255.255.240Remove
10.20.0.176/28 | 10.20.0.177 | 10.20.0.190 | 14 | 255.255.255.240Remove
10.20.0.192/28 | 10.20.0.193 | 10.20.0.206 | 14 | 255.255.255.240Remove
10.20.0.208/28 | 10.20.0.209 | 10.20.0.222 | 14 | 255.255.255.240Remove
10.20.0.224/28 | 10.20.0.225 | 10.20.0.238 | 14 | 255.255.255.240Remove
10.20.0.240/28 | 10.20.0.241 | 10.20.0.254 | 14 | 255.255.255.240
```

Just need 6 subnets correct?

xschildw: Looks good to me, maybe add an allowed pattern to make sure we input 3 octets as prefix since it's a change from normal template?

ahayden: That's a good idea @xschildw ... The subnets are different sizes @zaro0508; public ones are /28 and private ones are /26, so have 3 public and 3 private uses up addresses from 10.20.0.0->10.20.0.239

templates/vpc-mini.yaml
```diff
@@ -0,0 +1,450 @@
...
+Description: >-
+  Creates a VPC with public and private subnets and a managed NAT gateway.
+Parameters:
+  VpcName:
+    Description: The VPC name (i.e. Synapse-Prod)
+    Type: String
+  VpcSubnetPrefix:
+    Description: The VPC subnet prefix (i.e. 10.255.40)
+    Type: String
+    AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){2}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])$
```
ahayden, line 11: @xschildw this regex matches three octets, like `10.20.0`

templates/vpc-mini.yaml
zaro0508: i see now, looks fine then.  how about my question #1 from above?

templates/vpc-mini.yaml
```diff
@@ -0,0 +1,450 @@
...
+    Private:
+      CIDR: "16/26" #10.255.X.16-79
+    Public1:
+      CIDR: "80/28" #10.255.X.80-95
+    Private1:
+      CIDR: "96/26" #10.255.X.96-159
+    Public2:
+      CIDR: "160/28" #10.255.X.160-175
+    Private2:
+      CIDR: "176/26" #10.255.X.176-240
```
zaro0508, line 43: shouldn't the range be `10.255.X.176-239`?

templates/vpc-mini.yaml
ahayden: > 1. I think CIDR needs to be in X.X/YY notation?
The `CIDR` value in the map is appended to the `VpcSubnetPrefix` to make an IPv4 CIDR range... so the `CIDR` in the map can be `X/YY` is the `VpcSubnetPrefix` is `T.V.F` to make the usable CIDR string `T.V.F.X/YY`

templates/vpc-mini.yaml
ahayden: yes, you're right... will fix

PR #252, opened 2020-07-10
Use same sized ranges

ahayden:
The way I had defined subnets before was causing the ranges to overlap. This change makes every cidr block the same size and makes sure they don't overlap.


templates/vpc-mini.yaml
```diff
@@ -30,17 +30,17 @@ Mappings:
     VPC:
       CIDR: "0/24" #10.255.X.0/24 = 10.255.X.0-10.255.X.255
     Public:
-      CIDR: "0/28" #10.255.X.0-15
+      CIDR: "0/27" #10.255.X.0-31
```
zaro0508, line 33: the settings are correct but i think the comments are off by one. here's what i get from https://community.spiceworks.com/tools/subnet-calc/

Network | Start | End
------- | ------ | ------
10.255.20.0/27 | 10.255.20.1 | 10.255.20.30
10.255.20.32/27 | 10.255.20.33 | 10.255.20.62
10.255.20.64/27 | 10.255.20.65 | 10.255.20.94
10.255.20.96/27 | 10.255.20.97 | 10.255.20.126
10.255.20.128/27 | 10.255.20.129 | 10.255.20.158
10.255.20.160/27 | 10.255.20.161 | 10.255.20.190
10.255.20.192/27 | 10.255.20.193 | 10.255.20.222
10.255.20.224/27 | 10.255.20.225 | 10.255.20.254

ahayden: OK, updated

PR #253, opened 2020-07-16
[IT-876] Remove CFN custom resource

zaro0508:
The cfn-oidc-identity-provider custom resource has been moved
to https://github.com/Sage-Bionetworks-IT/cfn-cr-oidc-idp

We now can remove it from this repo.

depends on https://github.com/Sage-Bionetworks/scipool-infra/pull/147


ahayden:
Why does this only depend on a change to scicomp-infra? Won't all accounts need to be updated with the new location of the lambda source?

zaro0508:
Because we only use OIDC provider in scipooldev and scipoolprod accounts.  It's not installed on any other accounts.

PR #256, opened 2020-08-18
[IT-876] Delete s3objects macro

cascadianblue:
depends on Sage-Bionetworks/sageit-infra#31
depends on Sage-Bionetworks/sandbox-infra#148
depends on Sage-Bionetworks/scicomp-infra#223
depends on Sage-Bionetworks/scipool-infra#179
depends on Sage-Bionetworks/admincentral-infra#93
depends on Sage-Bionetworks/service-catalog-library#223


cascadianblue:
I might need some guidance on how to handle this in `templates/s3-bucket-v2.j2`:

```
# Add owner file to the synapse bucket, requires the cloudformation S3 objects macro
# https://github.com/Sage-Bionetworks/aws-infra/tree/master/lambdas/cfn-s3objects-macro
```

zaro0508:
> I might need some guidance on how to handle this in `templates/s3-bucket-v2.j2`:
> 
> ```
> # Add owner file to the synapse bucket, requires the cloudformation S3 objects macro
> # https://github.com/Sage-Bionetworks/aws-infra/tree/master/lambdas/cfn-s3objects-macro
> ```

That's just a comment so all you need to do with that is update the reference to the new repo "https://github.com/Sage-Bionetworks-IT/cfn-s3objects-macro"

cascadianblue:
I assumed that something after the comment somehow tied into that particular URL.

So if I understand correctly, I should just update the comment, right?

zaro0508:
> I assumed that something after the comment somehow tied into that particular URL.
> 
> So if I understand correctly, I should just update the comment, right?

that is correct @cascadianblue

PR #261, opened 2020-09-14
update travis to set AWS CLI env vars instead of using script

zaro0508:


.travis.yml
```diff
@@ -2,11 +2,11 @@ language: python
...
-  - pip install jsondiff cfn-lint taskcat==0.8.50
-  - wget https://github.com/Sage-Bionetworks/infra-utils/archive/master.zip -O /tmp/infra-utils.zip
-  - unzip -j -n /tmp/infra-utils.zip -x "infra-utils-master/.gitignore" "infra-utils-master/LICENSE" "infra-utils-master/*.md" "infra-utils-master/aws/*"
-  - ./setup_aws_cli.sh || travis_terminate 1
-  - pip install pre-commit aws-sam-cli
+  - pip install pre-commit jsondiff taskcat==0.8.50
+before_script:
+  - mkdir -p ~/.aws
+  - echo -e "[default]\nregion=us-east-1\nsource_profile=default\nrole_arn=$AwsCfServiceRoleArn" > ~/.aws/config
+  - echo -e "[default]\nregion=us-east-1\naws_access_key_id=$AwsTravisAccessKey\naws_secret_access_key=$AwsTravisSecretAccessKey" > ~/.aws/credentials
```
ahayden, line 9: Confirm this doesn't echo those vars to the console?

ahayden (approved):

.travis.yml
zaro0508: docs for it are here: https://docs.travis-ci.com/user/environment-variables/#defining-variables-in-repository-settings
```
By default, the value of these new environment variables is hidden from the export line in the logs. This corresponds to the behavior of encrypted variables in your .travis.yml. The variables are stored encrypted in our systems, and get decrypted when the build script is generated.

Similarly, we do not provide these values to untrusted builds, triggered by pull requests from another repository.
```

PR #263, opened 2020-09-23
SSM parameters template

zaro0508:
Add a cloudformation template to install parameters to the
SSM parameter store.  This is a generic jinja template that can
be used to install multiple SSM parameters into any AWS account.
This is only for non-securestring parameters.  Cloudformation
does not support adding securestring parameters into the SSM.

### DIFF
--- a/templates/ssm-parameters.j2
+++ b/templates/ssm-parameters.j2
@@ -1,0 +1,46 @@
+# This is a cloudformation jinjaized template to install parameters into the
+# SSM parameter store.  This is only for non-securestring parameters.
+# Cloudformation does not support adding securestring parameters into the SSM.
+#
+# Usage:
+#  Create a sceptre template `my-ssm-params.yaml`
+#
+#  template_path: remote/ssm-parameters.j2
+#  hooks:
+#    before_launch:
+#      - !cmd "curl https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/master/ssm-parameters.j2 --create-dirs -o templates/remote/ssm-parameters.j2"
+#  stack_name: my-ssm-vars
+#  sceptre_user_data:
+#    Prefix: /my-ssm-vars/
+#    Parameters:
+#      - Name: param1
+#        Value: foo
+#      - Name: param2
+#        Value: foo
+
+Description: "Setup parameters in the SSM parameter store"
+AWSTemplateFormatVersion: "2010-09-09"
+Resources:
+  {% for parameter in sceptre_user_data.Parameters %}
+  {%- set name = parameter.Name -%}
+  {%- set value = parameter.Value -%}
+    {{ name }}:
+      Type: AWS::SSM::Parameter
+      Properties:
+        Name: {{ sceptre_user_data.Prefix }}{{ name }}
+        Value: {{ value }}
+        Type: 'String'
+  {% endfor %}
+Outputs:
+  Prefix:
+    Value: {{ sceptre_user_data.Prefix }}
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-Prefix'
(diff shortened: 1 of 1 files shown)


templates/ssm-parameters.j2
```diff
@@ -0,0 +1,46 @@
...
+#    Parameters:
+#      - Name: param1
+#        Value: foo
+#      - Name: param2
+#        Value: foo
+
+Description: "Setup parameters in the SSM parameter store"
+AWSTemplateFormatVersion: "2010-09-09"
+Resources:
+  {% for parameter in sceptre_user_data.Parameters %}
```
cascadianblue, line 24: I don't have an easy way to test this, but it looks like this might create an extra newline where you don't want it. Would it make sense to change either the `{%` to `{%-`, or `%}` to `-%}` ?

Minor issue either way I guess

templates/ssm-parameters.j2
```diff
@@ -0,0 +1,46 @@
...
+  {%- set name = parameter.Name -%}
+  {%- set value = parameter.Value -%}
+    {{ name }}:
+      Type: AWS::SSM::Parameter
+      Properties:
+        Name: {{ sceptre_user_data.Prefix }}{{ name }}
```
cascadianblue, line 30: would it be more consistent stylistically to add a line `{%- set prefix = sceptre_user_data.Prefix -%}` (outside the for-loop) and then change this to `Name: {{ prefix }}{{ name }}`?

cascadianblue (commented):
Didn't we decide a while ago that we want YAML parameters (values) to be in quotes whenever possible in sceptre templates?

templates/ssm-parameters.j2
cascadianblue: Or even better: `Name: "{{ prefix }}{{ name }}"` ?

templates/ssm-parameters.j2
zaro0508: Why would you think this would create an extra newline?  The purpose of this PR is to move this template, which is already working in production, to a new location.  take a look at https://github.com/Sage-Bionetworks/synapse-login-aws-infra/pull/42

templates/ssm-parameters.j2
zaro0508: no that won't work, it needs to be `sceptre_user_data.Prefix`.  again, template is only being moved, it's already working as is.

cascadianblue:
> Didn't we decide a while ago that we want YAML parameters (values) to be in quotes whenever possible in sceptre templates?

Just kidding, I remembered that we lint the configs this way, but not the templates

templates/ssm-parameters.j2
cascadianblue: Not sure I understand why that wouldn't work, but my comment was just stylistic anyway.

cascadianblue (approved):

templates/ssm-parameters.j2
zaro0508: it wouldn't work because `Prefix` is not in the `sceptre_user_data.Parameters` dict.  jinja requires a full reference to `Prefix` so it needs to be `sceptre_user_data.Prefix`.

ahayden (approved):
Are SSM parameters removed when a stack is replaced? Would it be useful to document a naming convention for paths in the parameter store? Running this template will overwrite any parameters already in the store, which could potentially be bad.

zaro0508:
> Are SSM parameters removed when a stack is replaced? Would it be useful to document a naming convention for paths in the parameter store? Running this template will overwrite any parameters already in the store, which could potentially be bad.

Yes, running this template could potentially overwrite SSM parameter store values.  I'm not sure how we would prevent that even with documentation and a naming convention.  I'm open to suggestions, but even better yet, you can write the docs, somewhere, to help not let this happen.   I guess i would say that code review is what he have to mitigate this potential problem.

PR #266, opened 2020-11-24
[IT-1004] Setup transit gateway spokes

zaro0508:
Add a template to setup transit gateway on the spoke end.
This template depends on a shared with ID `TransitGatewayId`.

Will require the following workflow:
1. share transit VPC (in transit account) to spoke accounts
2. accept sharing invitation in spoke account
3. run transit-gateway-spoke.yaml template to setup the spoke
account.


ahayden (approved):
Looks good

PR #267, opened 2020-11-28
[IT-1006] add TGW security group

zaro0508:
Add a transit gateway security group to allow access to spoke VPCs from
the the HUB VPC.  The idea is to apply this SG to resources in spoke
accounts so that users can access the resource when connected to the AWS
client VPN.


xschildw (approved):
So we'll add this group in existing templates to add the SG to existing resources?

zaro0508:
yes, that is correct @xschildw

PR #268, opened 2020-12-02
[IT-1005] template to create friendly redirect URL

xschildw:
[IT-1005] This template creates an S3 website configured with a routing rule to redirect to another website, Cloudfront is setup to handle https. Also supports an optional key to replace the incoming one. First intended use is a friendly url to the self-service portal for VPN client.

[IT-1005]: https://sagebionetworks.jira.com/browse/IT-1005

### DIFF
--- a/templates/s3-redirector.yaml
+++ b/templates/s3-redirector.yaml
@@ -1,0 +1,103 @@
+# This template sets up an S3 static website with a routing rule to redirect (307)
+# to another website, optionally replacing the key. Also sets up a Cloudfront
+# distribution to handle https.
+AWSTemplateFormatVersion: 2010-09-09
+Description: >-
+  Provision a S3 static website redirector with Cloudfront
+Parameters:
+  SourceHostName:
+    Description: The source endpoint (to redirect from)
+    Type: String
+    AllowedPattern: (?!-)[a-zA-Z0-9-.]{1,63}(?<!-)
+    ConstraintDescription: must be a valid DNS zone name.
+  SourceAcmCertificateArn:
+    Description: The Amazon Resource Name (ARN) of an AWS Certificate Manager (ACM) certificate.
+    Type: String
+    AllowedPattern: "arn:aws:acm:.*"
+    ConstraintDescription: must be a valid certificate ARN.
+  TargetHostName:
+    Type: String
+    Description: The target endpoint 
+    AllowedPattern: (?!-)[a-zA-Z0-9-.]{1,63}(?<!-)
+    ConstraintDescription: must be a valid DNS zone name.
+  TargetKey:
+    Type: String
+    Description: Optional target key
+    Default: ""
+Conditions:
+  HasTargetKey: !Not [ !Equals ["", !Ref TargetKey ] ]
+Resources:
+  WebsiteBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: PublicRead
+      BucketName: !Ref SourceHostName
+      WebsiteConfiguration:
+        IndexDocument: index.html
+        ErrorDocument: error.html
+      RoutingRules:
+        RedirectRule:
+          HostName: !Ref TargetHostName
+          HttpRedirectCode: '307'
+          Protocol: https
+          ReplaceKeyWith: !If [ !Ref HasTargetKey, !Ref TargetKey, !Ref 'AWS::NoValue' ]
+     DeletionPolicy: Retain
+     UpdateReplacePolicy: Retain
+
+  Cloudfront:
+    Type: AWS::CloudFront::Distribution
+    Properties:
+      DistributionConfig:
+        Comment: Cloudfront Distribution pointing to S3 bucket
+        Aliases:
+          - !Ref SourceHostName
+        Enabled: true
+        Origins:
+          - DomainName: !Select [2, !Split ["/", !GetAtt WebsiteBucket.WebsiteURL]]
+            Id: S3Origin
+            CustomOriginConfig:
+              HTTPPort: 80
+              HTTPSPort: 443
+              OriginProtocolPolicy: http-only
+        HttpVersion: 'http2'
+        DefaultRootObject: index.html
+        DefaultCacheBehavior:
+          DefaultTTL: 3600
+          AllowedMethods:
+            - GET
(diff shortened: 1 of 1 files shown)


templates/s3-redirector.yaml
```diff
@@ -0,0 +1,103 @@
...
+Conditions:
+  HasTargetKey: !Not [ !Equals [ "", !Ref TargetKey ] ]
+Resources:
+  WebsiteBucket:
+    Type: AWS::S3::Bucket
+    Properties:
+      AccessControl: PublicRead
+      BucketName: !Ref SourceHostName
+      WebsiteConfiguration:
+        IndexDocument: index.html
```
zaro0508, line 35: does this template create an `index.html` file or is this something we need to manually add to the bucket?

templates/s3-redirector.yaml
```diff
@@ -0,0 +1,103 @@
...
+    AllowedPattern: "arn:aws:acm:.*"
+    ConstraintDescription: must be a valid certificate ARN.
+  TargetHostName:
+    Type: String
+    Description: The target endpoint
+    AllowedPattern: (?!-)[a-zA-Z0-9-.]{1,63}(?<!-)
+    ConstraintDescription: must be a valid DNS zone name.
+  TargetKey:
+    Type: String
+    Description: Optional target key
```
zaro0508, line 24: Could you please add a little bit more info on what this is for?

templates/s3-redirector.yaml
```diff
@@ -0,0 +1,103 @@
...
+  SourceAcmCertificateArn:
+    Description: The Amazon Resource Name (ARN) of an AWS Certificate Manager (ACM) certificate.
+    Type: String
+    AllowedPattern: "arn:aws:acm:.*"
+    ConstraintDescription: must be a valid certificate ARN.
+  TargetHostName:
+    Type: String
+    Description: The target endpoint
+    AllowedPattern: (?!-)[a-zA-Z0-9-.]{1,63}(?<!-)
+    ConstraintDescription: must be a valid DNS zone name.
```
zaro0508, line 21: so this is the VPN SSP endpoint.  does this mean we need to also create a hosted zone for it?

zaro0508 (commented):
wow, we need cloudfront just to setup a redirect?  setting up a redirect is way more complicated than i anticipated. 

This generally looks good, just had a few question before i approve

templates/s3-redirector.yaml
xschildw: No, I don't think we need the actual file. The redirect should kick in before we try to access it.

templates/s3-redirector.yaml
xschildw: My understanding is that it will reside in the sageit.org domain so all we need to do is add a CNAME that will point to the URL of the Cloudfront distribution (btw to answer your first comment, we only need Cloudfront to handle https).
My thinking was to keep this one as simple as possible (there are cases where the CNAME already exists and just needs to be updated, and not always created with Cloudformation). For our application, the plan was to create a separate template to create the CNAME pointing to the CloudfrontEndpoint of this template. Come to think of it, maybe I should make the CNAME creation optional in this one (i.e. create if specified)?

templates/s3-redirector.yaml
xschildw: Will do. Thx!

zaro0508 (changes requested):

templates/s3-redirector.yaml
```diff
@@ -14,17 +14,25 @@ Parameters:
...
     AllowedPattern: "arn:aws:acm:.*"
     ConstraintDescription: must be a valid certificate ARN.
-  TargetHostName:
+  SourceHostedZoneId:
+    Description: The ID of the hosted zone where the source endpoint will be created. Needs to be in same account.
     Type: String
+    MaxLength: 32
+    AllowedPattern: "[A-Z0-9]*"
+    ConstraintDescription: Must be a valid hosted zone ID
+    Default: ""
```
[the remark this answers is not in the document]
xschildw: The zone may or may not have been created with a template (for this application, I don't think sageit.org has). If it has we should be be able to pass a sceptre stack output else just the ID of a hosted zone in same account. In other cases the zone is outside the account, so I made it optional.

templates/s3-redirector.yaml
```diff
@@ -0,0 +1,123 @@
...
+        ErrorDocument: error.html
+        RoutingRules:
+          -
+            RedirectRule:
+              HostName: !Ref TargetHostName
+              HttpRedirectCode: '307'
+              Protocol: https
+              ReplaceKeyWith: !If [ HasTargetKey, !Ref TargetKey, !Ref "AWS::NoValue" ]
+    DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
```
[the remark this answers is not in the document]
xschildw: But if the bucket goes so does the redirect rule, thought it should be protected like a regular bucket.

templates/s3-redirector.yaml
```diff
@@ -0,0 +1,123 @@
...
+          TargetOriginId: S3Origin
+          ForwardedValues:
+            QueryString: true
+            Cookies:
+              Forward: none
+          ViewerProtocolPolicy: redirect-to-https
+        PriceClass: PriceClass_100
+        ViewerCertificate:
+          AcmCertificateArn: !Ref SourceAcmCertificateArn
+          SslSupportMethod: sni-only
```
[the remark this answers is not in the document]
xschildw: OK

templates/s3-redirector.yaml
```diff
@@ -0,0 +1,123 @@
...
+        Aliases:
+          - !Ref SourceHostName
+        Enabled: true
+        Origins:
+          - DomainName: !Select [2, !Split ["/", !GetAtt WebsiteBucket.WebsiteURL]]
+            Id: S3Origin
+            CustomOriginConfig:
+              HTTPPort: 80
+              HTTPSPort: 443
+              OriginProtocolPolicy: http-only
```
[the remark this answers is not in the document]
xschildw: As I recall when the origin is an S3 static website it can only be http, and we can't use S3Origin because the routing don't work in that case. Will double-check.

PR #273, opened 2021-02-25
Increase complexity of inputs

ahayden:
This increases the requirements of IAM user passwords, which should be fine as they are becoming less and less commonly used.


xschildw (approved):

zaro0508 (approved):
looks good.  not sure if you want to customize per account but you can do that as well by passing in params.  here's an example: https://github.com/org-formation/aws-resource-providers/blob/master/iam/password-policy/example.yml

PR #275, opened 2021-03-04
service account template

zaro0508:
Instead of having two separate service account cfn templates, one for managed policies
and another one for custom policies we create a singlecfn template that will handle
both types of policies. User can provide both or they can provide either
one.  User must provide at least one type of policy though.


xschildw (approved):

templates/IAM/service-account.yaml
```diff
@@ -1,20 +1,62 @@
 Description: Setup an AWS service account and role
 AWSTemplateFormatVersion: 2010-09-09
 Parameters:
-  ServiceUserPolicies:
```
xschildw, line 4: Renaming the parameter is gonna break back-compat, problem?

zaro0508: i pin it to a tag that has the old version.  https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/88

PR #278, opened 2021-03-12
add sts:TagSession to service role

zaro0508:


zaro0508:
replaced this wacko PR with #279

PR #280, opened 2021-03-12
add SCPs

zaro0508:
Add service control policies (SCP) that will be deployed by org-formation
into an organizations account.


ahayden (commented):
Good idea

PR #284, opened 2021-03-19
[IT-1121] Update synapse bucket

zaro0508:
When using the AWS CLI it's required to pass `--acl bucket-owner-full-control`
when uploading files to synapse bucket

### DIFF
--- a/templates/s3-bucket-v2.j2
+++ b/templates/s3-bucket-v2.j2
@@ -105,6 +105,9 @@ Resources:
             AllowedOrigins: ['*']
             AllowedMethods: [GET, POST, PUT, HEAD]
             MaxAge: 3000
+      OwnershipControls:
+        Rules:
+          - ObjectOwnership: BucketOwnerPreferred
       LifecycleConfiguration:
         Rules:
         - Id: DataLifecycleRule
@@ -175,6 +178,16 @@ Resources:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
             Resource: !If [EnableEncryption, !GetAtt SynapseEncryptedExternalBucket.Arn, !GetAtt SynapseExternalBucket.Arn]
+          -
+            Sid: "RequireCanonicalIdOnObjectUpdates"
+            Effect: Deny
+            Principal:
+              AWS: !Ref GrantAccess
+            Action: s3:PutObject
+            Resource: !If [EnableEncryption, !Sub "${SynapseEncryptedExternalBucket.Arn}/*", !Sub "${SynapseExternalBucket.Arn}/*"]
+            Condition:
+              StringNotEquals:
+                s3:x-amz-acl: bucket-owner-full-control
           -
             Sid: "WriteAccess"
             Effect: "Allow"


brucehoff (approved):

BrunoGrandePhD:
I don't have any context on this particular repository. Which S3 buckets would be affected by this PR?

zaro0508:
> I don't have any context on this particular repository. Which S3 buckets would be affected by this PR?

all buckets that have been provisioned by the Sage-Bionetworks/scicomp-provisioner and Sage-Bionetworks/sandbox-provisioner repos.

BrunoGrandePhD:
Thanks for the clarification! So, would this automatically update all existing buckets with this new policy? Or would it only apply to new buckets going forward?

I ask because **[as promised](https://sagebionetworks.slack.com/archives/CEFQD0KU1/p1616001020035200?thread_ts=1616000394.034300&cid=CEFQD0KU1)**, I'm drafting an email to owners of S3 buckets in `scicomp` and I would like to know whether they need to take any action.

zaro0508:
It will update all buckets provisioned with the [s3-bucket-v2.j2 template from aws-infra repo](https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/config/prod/TESLA-download-bucket.yaml#L47).  htan buckets are not provisioned with that template they use a template from the [s3-synapse-sync repo](https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/config/prod/htan-dcc-ohsu.yaml#L11).

I just noticed that there are buckets that were provisioned with the older [s3-bucket.yaml template](https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/config/prod/AMP-Emory-Bucket.yaml#L25) as well so we'll need to update the policies in that template as well.

so to answer your question @BrunoGrandePhD  it depends on which template was used to provision buckets.  We basically need to update all the templates that were used.

BrunoGrandePhD:
Only one of the S3 buckets is provisioned using `s3-bucket-v2.j2`. If you plan on updating the older `s3-bucket.yaml` template (not sure about `managed-s3WebCloudfront.yaml`), then we should alert the bucket owners about the change. 

```
AMP-Emory-Bucket.yaml                 ampadportal@sagebase.org        s3-bucket.yaml
JKG-SciComp-S3.yaml                   jake.gockley@sagebase.org       s3-bucket.yaml
TESLA-download-bucket.yaml            james.eddy@sagebase.org         s3-bucket-v2.j2
aacr-interactome-s3website-new.yaml   xindi.guo@sagebase.org          managed-s3WebCloudfront.yaml
amp-mayo-sinai.yaml                   william.poehlman@sagebase.org   s3-bucket.yaml
gates-ki-001.yaml                     larsson.omberg@sagebase.org     s3-bucket.yaml
mep-lincs-s3.yaml                     larsson.omberg@sagebase.org     s3-bucket.yaml
usage-statistics-S3.yaml              larsson.omberg@sagebase.org     s3-bucket.yaml
veoibd-s3.yaml                        cindy.molitor@sagebase.org      s3-bucket.yaml
```

I'll send out the email in a moment and copy you.

ahayden (approved):
lgtm

templates/s3-bucket-v2.j2
```diff
@@ -175,6 +178,16 @@ Resources:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
             Resource: !If [EnableEncryption, !GetAtt SynapseEncryptedExternalBucket.Arn, !GetAtt SynapseExternalBucket.Arn]
+          -
+            Sid: "RequireCanonicalIdOnObjectUpdates"
+            Effect: Deny
+            Principal:
+              AWS: !Ref GrantAccess
```
brucehoff, line 185: @zaro0508  I think this logic should be changed.  We should _not_ deny `PutObject` access to _all_ in the 'principals' list if they don't add the `bucket-owner-full-control` ACL.  Instead we should only apply this condition to principals who are **not** in the account that owns the bucket.  I.e., if a principal in the account that owns the bucket uploads a file without the ACL that is OK because the bucket owner *already* owns the uploaded object.    This change is important because, as @BrunoGrandePhD discovered for us, when uploading through the S3 console there is no way to add an ACL to the created object.

brucehoff (changes requested):

templates/s3-bucket-v2.j2
zaro0508: I think i have done what is requested.  I used an example from aws docs to come up with the configs,  https://aws.amazon.com/premiumsupport/knowledge-center/explicit-deny-principal-elements-s3/

brucehoff: It looks like the change you made addresses the second issue described above (omitting Synapse from the constraint) but does not address the first issue (omitting principals in the account that owns the bucket).  Do you agree?

zaro0508: Done.  The AWS account owns the bucket.  Cloudformation does not have a way to reference account IDs therefore the user will need to pass the ID in.  Info on how to get the user ID for an AWS account is here https://docs.aws.amazon.com/AmazonS3/latest/userguide/finding-canonical-user-id.html  it will look something like this..
```
➜   aws s3api list-buckets --query Owner.ID --output text
9038e06f22b4c2611873a9ac491ce754aa2353b45e19ab508577ee99863128ed
```

BrunoGrandePhD (approved):
Based on my little experience with CloudFormation, I see nothing wrong here. 

🚀

brucehoff (approved):

PR #286, opened 2021-03-25
[IT-1101] add template to setup aws budgets

zaro0508:
This cfng template will allow us to use org-formation to configure
a budget for each AWS account and get notifications when budgets
are exceeding expectations.

### DIFF
--- a/templates/Billing/budgets.yaml
+++ b/templates/Billing/budgets.yaml
@@ -1,0 +1,90 @@
+AWSTemplateFormatVersion: '2010-09-09'
+
+Parameters:
+
+  budgetName:
+    Type: String
+
+  emailAddress:
+    Type: String
+    Default: ''
+
+  budgetAmount:
+    Type: Number
+    Default: 50
+
+Conditions:
+
+  hasSubscription: !Not [ !Equals [ '', !Ref emailAddress ] ]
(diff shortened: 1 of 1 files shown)


templates/Billing/budgets.yaml
```diff
@@ -0,0 +1,90 @@
...
+            Action: sns:Publish
+            Resource: !Ref NotificationTopic
+
+  # note that changing some attributes on Budget resources will result in an error.
+  # this unfortunately is a cloudformation error described here: https://github.com/aws-cloudformation/aws-cloudformation-coverage-roadmap/issues/365
+  # workaround is to delete the budget resource and then re-add it.
+  Budget:
+    Type: AWS::Budgets::Budget
+    Properties:
+      Budget:
```
ahayden, line 57: Does the budget apply to all resources in the account by default?

xschildw (approved):
Sounds good to me, wondering if we need 80% actual (if forecasting is any good).

ahayden (approved):
Looks good

templates/Billing/budgets.yaml
zaro0508: it applies to the entire account so yes.

PR #288, opened 2021-03-26
[IT-1101] Add templates to setup AWS master member guardduty

zaro0508:
These templates will allow org-formation to deploy guardduty in a
master member configuration[1].  All templates are just copied
from the ofn reference project[2]

[1] https://docs.aws.amazon.com/guardduty/latest/ug/guardduty_accounts.html
[2] https://github.com/org-formation/org-formation-reference/tree/master/src/templates/070-guard-duty

### DIFF
--- a/templates/GuardDuty/trusted-ips-bucket.yaml
+++ b/templates/GuardDuty/trusted-ips-bucket.yaml
@@ -1,6 +1,11 @@
 # From https://github.com/org-formation/org-formation-reference/tree/master/src/templates/070-guard-duty
 AWSTemplateFormatVersion: '2010-09-09'
 
+Metadata:
+  cfn-lint:
+    config:
+      ignore_checks: [W2001]
+
 Parameters:
   resourcePrefix:
     Type: String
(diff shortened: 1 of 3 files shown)


templates/GuardDuty/trusted-ips-bucket.yaml
```diff
@@ -0,0 +1,53 @@
...
+      PolicyDocument:
+        Version: 2012-10-17
+        Statement:
+          - Sid: GuardDutyRead
+            Effect: Allow
+            Principal:
+              Service: guardduty.amazonaws.com
+            Action:
+              - s3:GetObject
+            Resource: !Sub '${TrustedIpsBucket.Arn}/trusted_ips.txt'
```
ahayden, line 47: Maybe it doesn't matter if the trusted IPs list is empty... but do want to have a list like this?

zaro0508: do you have a concern with it being here?  if not then lets just add as an optional feature.  if you have a concern later you are free to remove it if you want.

ahayden (approved):
All looks good. Where would we subscribe to the notification topics? IE to get an email

zaro0508:
> All looks good. Where would we subscribe to the notification topics? IE to get an email

it's here @ahayden: 
https://github.com/Sage-Bionetworks/aws-infra/pull/288/files#diff-afc5705ba1fe086bdf9b9ab4c0f51b261dc6ed4e07c0119eeb54c75d5609e9b8R29

PR #289, opened 2021-03-26
[T-1101] Add templates to setup AWS master member config

zaro0508:
Add templates to allow org-formation to deploy aws config in a
master member configuration[1].  This emplate is just copied
from the ofn reference project[2]

[1] https://docs.aws.amazon.com/organizations/latest/userguide/services-that-can-integrate-config.html
[2] https://github.com/org-formation/org-formation-reference/blob/master/src/templates/080-aws-config-inventory

### DIFF
--- a/templates/Config/config.yaml
+++ b/templates/Config/config.yaml
@@ -1,6 +1,8 @@
 # From https://github.com/org-formation/org-formation-reference/blob/master/src/templates/080-aws-config-inventory/config.yml
 AWSTemplateFormatVersion: '2010-09-09'
 
+# This is an org-formation file, not a cloudformation file therefore some cfn-lint rules do not apply
+# rules reference: https://github.com/aws-cloudformation/cfn-python-lint/blob/master/docs/rules.md#rules-1
 Metadata:
   cfn-lint:
     config:
@@ -97,7 +99,7 @@ Resources:
       AssumeRolePolicyDocument:
         Version: '2012-10-17'
         Statement:
-        - Sid: AssumeRole1
+        - Sid: ConfigServiceAssumeRole
           Effect: Allow
           Principal:
             Service: 'config.amazonaws.com'
--- a/templates/Config/config.yaml
+++ b/templates/Config/config.yaml
@@ -1,0 +1,118 @@
+# From https://github.com/org-formation/org-formation-reference/blob/master/src/templates/080-aws-config-inventory/config.yml
+AWSTemplateFormatVersion: '2010-09-09'
+
+Metadata:
+  cfn-lint:
+    config:
+      ignore_checks: [W2001]
+
+Parameters:
+  resourcePrefix:
+    Type: String
+
+  bucketName:
+    Type: String
+    Description: 'Name of the central S3 bucket containing AWS Config audit findings'
+
+Resources:
+  ConfigAuditBucket:
+    OrganizationBinding: !Ref LogArchiveBinding
+    DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
+    Type: 'AWS::S3::Bucket'
+    Metadata:
+      cfn-lint:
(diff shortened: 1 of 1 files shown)


templates/Config/config.yaml
```diff
@@ -0,0 +1,118 @@
...
+          - ServerSideEncryptionByDefault:
+              SSEAlgorithm: AES256
+
+  ConfigAuditBucketPolicy:
+    OrganizationBinding: !Ref LogArchiveBinding
+    Type: AWS::S3::BucketPolicy
+    Metadata:
+      cfn-lint:
+        config:
+          ignore_checks: [E1012, E3001]
```
ahayden, line 46: Can you include a comment about what these checks are ignoring?

templates/Config/config.yaml
```diff
@@ -0,0 +1,118 @@
...
+Resources:
+  ConfigAuditBucket:
+    OrganizationBinding: !Ref LogArchiveBinding
+    DeletionPolicy: Retain
+    UpdateReplacePolicy: Retain
+    Type: 'AWS::S3::Bucket'
+    Metadata:
+      cfn-lint:
+        config:
+          ignore_checks: [E1012, E3001]
```
ahayden, line 26: Can you include a comment about what these checks are ignoring?

templates/Config/config.yaml
```diff
@@ -0,0 +1,118 @@
+# From https://github.com/org-formation/org-formation-reference/blob/master/src/templates/080-aws-config-inventory/config.yml
+AWSTemplateFormatVersion: '2010-09-09'
+
+Metadata:
+  cfn-lint:
+    config:
+      ignore_checks: [W2001]
```
ahayden, line 7: Can you include a comment about what these checks are ignoring?

templates/Config/config.yaml
```diff
@@ -0,0 +1,118 @@
...
+      RecordingGroup:
+        AllSupported: true
+        IncludeGlobalResourceTypes: true
+      RoleARN: !GetAtt ConfigurationRecorderRole.Arn
+
+  DeliveryChannel:
+    Type: 'AWS::Config::DeliveryChannel'
+    Properties:
+      ConfigSnapshotDeliveryProperties:
+        DeliveryFrequency: One_Hour
```
ahayden, line 89: This is the fastest option, looks like.

templates/Config/config.yaml
```diff
@@ -0,0 +1,118 @@
...
+
+  ConfigurationRecorderRole:
+    Type: 'AWS::IAM::Role'
+    Properties:
+      ManagedPolicyArns:
+      - 'arn:aws:iam::aws:policy/service-role/AWSConfigRole'
+      AssumeRolePolicyDocument:
+        Version: '2012-10-17'
+        Statement:
+        - Sid: AssumeRole1
```
ahayden, line 100: This should have a name that explains its purpose

ahayden (changes requested):
I requested a few small explanations in the template

templates/Config/config.yaml
zaro0508: added a general comment for these ignore above this section.

templates/Config/config.yaml
zaro0508: i think general comment above covers it.

templates/Config/config.yaml
zaro0508: i think general comment above covers it.

templates/Config/config.yaml
zaro0508: updated, but really i just got this file from https://github.com/org-formation/org-formation-reference/blob/master/src/templates/080-aws-config-inventory/config.yml

ahayden (approved):
Looks fine, thanks!

PR #290, opened 2021-03-26
[IT-1101] remove cloudtrail resources

zaro0508:
We are replacing the sceptre deployment of cloudtrail with org-formation
deployment.  The essentials.yaml template is deployed with sceptre so
removing cloudtrail resources from the template.  Keep the bucket so
we can retail the old trail data.


xschildw (approved):

ahayden (approved):
Looks good. Confirmed CloudTrail logging is already updated with this change: https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/117

PR #291, opened 2021-04-10
IT-1187: Parameterize minPasswordLenght with default of 16

xschildw:
The new default of 18 does not work with the initial password setup in ssm. Set to 16 as default overridable in _parameters.yaml.

### DIFF
--- a/templates/ORG/password-policy.yaml
+++ b/templates/ORG/password-policy.yaml
@@ -1,10 +1,17 @@
 AWSTemplateFormatVersion: '2010-09-09'
 
+Parameters:
+  MinPasswordLength:
+    Default: 16
+    Description: The minimum length of a password
+    Type: Number
+    MinValue: 16
+
 Resources:
   PasswordPolicy:
     Type: Community::IAM::PasswordPolicy
     Properties:
-      MinimumPasswordLength: 18
+      MinimumPasswordLength: !Ref MinPasswordLength
       RequireLowercaseCharacters: true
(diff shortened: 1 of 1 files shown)


templates/ORG/password-policy.yaml
```diff
@@ -1,10 +1,17 @@
 AWSTemplateFormatVersion: '2010-09-09'
 
+Parameters:
+  MinPasswordLength:
+    Default: 16
+    Description: The minimum length of a password
+    Type: Number
+    MinValue: 16
```
zaro0508, line 8: Min is 16?  shouldn't this be a range that like min of 8 and max of 32 or something?

zaro0508 (approved):
approve with suggestion

templates/ORG/password-policy.yaml
xschildw: Sure, I just put it to 16 to make sure we would not be able to change it to anything less than current min but then again we can change everything so...

zaro0508: i see the problem @xschildw.  we can fix it in one of two ways, 
1) set MinPasswordLength to 8 for synapsedev account while keeping the others at 18 or 
2) change the default password in SSM in synapsedev to some value that’s 18 chars long.

i can do #2 for you if you like.

ahayden: I like option (2) best...

xschildw:
We're going ahead with option 2.

PR #293, opened 2021-04-19
Create a medium-sized mostly private VPC of ~4000 hosts

tthyer:
The purpose of this PR is to create a medium-sized VPC consisting mostly of private subnets. There is a public subnet only to accommodate the NAT gateway. There should be no public ingress.

This VPC will be used in the NextFlow account(s).


tthyer:
I cannot set reviewers on this repo, but would like for this PR to be reviewed by @ahayden @xschildw @zaro0508 and  @BrunoGrandePhD.

BrunoGrandePhD:
I defer to others for this PR. I don't have sufficient expertise on VPCs and networking to give a sensible review.

tthyer:
@zaro0508 I have set this up in the dev account and used it to create a NF compute environment, but there is an issue -- workflows hang when I use that compute environment (the head job is never created). Another compute environment that uses the default VPC works just fine. Before I go back to NF and ask them for more guidance, would like for you to take a look and see whether this VPC is generally what you had in mind, or if I've cut too much.

templates/VPC/vpc-20-private.yaml
```diff
@@ -0,0 +1,396 @@
...
+    Type: "AWS::EC2::VPC"
+    Properties:
+      EnableDnsSupport: true
+      EnableDnsHostnames: true
+      CidrBlock: !Join
+        - '.'
+        - - !Ref VpcSubnetPrefix
+          - !FindInMap [SubnetConfig, VPC, CIDR]
+      Tags:
+        - Key: "Name"
```
zaro0508, line 54: you can probably remove all tags from this templates.  it's better to pass them in with sceptre `stack_tags:`.

templates/VPC/vpc-20-private.yaml
```diff
@@ -0,0 +1,396 @@
+AWSTemplateFormatVersion: "2010-09-09"
+Description: >-
+  Creates a VPC with public and private subnets and a managed NAT gateway.
```
zaro0508, line 3: update with more specific description?

templates/VPC/vpc-20-private.yaml
```diff
@@ -0,0 +1,396 @@
...
+          - !FindInMap [SubnetConfig, VPC, CIDR]
+      Tags:
+        - Key: "Name"
+          Value: !Ref VpcName
+  PublicSubnet:
+    Type: "AWS::EC2::Subnet"
+    Properties:
+      MapPublicIpOnLaunch: true
+      VpcId: !Ref VPC
+      CidrBlock: !Join
```
zaro0508, line 61: minor, but i would really like to see all these `!Join` converted to `!Sub`

templates/VPC/vpc-20-private.yaml
```diff
@@ -0,0 +1,396 @@
...
+          IpProtocol: "-1"
+          Description: "Allow all VPN traffic"
+      # CF does not support removing all rules, workaround is to add a pointless rule
+      SecurityGroupEgress:
+        - CidrIp: "0.0.0.0/0"
+          FromPort: -1
+          ToPort: -1
+          IpProtocol: "-1"
+  # Allow access to bastian hosts
+  BastianSecurityGroup:
```
zaro0508, line 290: i don't think this is necessary since all instances will be in private subnet?

templates/VPC/vpc-20-private.yaml
```diff
@@ -0,0 +1,396 @@
...
+          FromPort: 22
+          ToPort: 22
+          IpProtocol: tcp
+          Description: "Allow SSH connection"
+      SecurityGroupEgress:
+        - CidrIp: "0.0.0.0/0"
+          FromPort: -1
+          ToPort: -1
+          IpProtocol: "-1"
+  QuarantineSecurityGroup:
```
zaro0508, line 307: same as BastianSecurityGroup

zaro0508 (changes requested):
i think you are missing a `InboundHTTPPublicNetworkAclEntry`, https://github.com/Sage-Bionetworks/aws-infra/blob/807c7f4547d9ef09605e94bf7ec6df412599ac72/templates/VPC/vpc-mini.yaml#L203.  Maybe that's why it's not working?  Otherwise it may just depend on how nextflow works.  Maybe it wants multiple public subnets? dunno.

templates/VPC/vpc-20-private.yaml
```diff
@@ -0,0 +1,396 @@
...
+    Type: String
+    Default: "10.1.0.0/16"
+    AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(1[6-9]|2[0-8]))$
+    ConstraintDescription: CIDR block parameter must be in the form x.x.x.x/16-28
+Mappings:
+  SubnetConfig:
+    VPC:
+      CIDR: "0.0/20"  # 10.255.0.0 - 10.255.15.255
+    Private:
+      CIDR: "0.0/22"  # 10.255.0.0 - 10.255.3.255
```
zaro0508, line 32: my preference is to split a /20 CIDR into 4 equal subnets, https://pasteboard.co/JZ8SNxK.png

tthyer: @zaro0508, I agree that would be tidier and preferable, but according to the AWS docs a public subnet is necessary for a NAT gateway. @ahayden suggested we go with 4096 addresses (`/20`), and I split them in this way to not waste too many on the public subnet. Alternately, perhaps I should go to  `/19`?

templates/VPC/vpc-20-private.yaml
tthyer: Agree but this a bit time consuming and needs to be done throughout a lot of aws-infra templates -- they're showing up here because this stuff was copied and pasted from another VPC template. Don't you have a new intern yet? 😁

zaro0508: OMG! We keep asking @brucehoff but he's decided to allocate interns to other projects 😞

templates/VPC/vpc-20-private.yaml
zaro0508: sounds good, however there might be a one off error in your comment of ip ranges? https://pasteboard.co/JZ9vjzB.png

templates/VPC/vpc-20-private.yaml
tthyer: removed, except for naming the VPC

templates/VPC/vpc-20-private.yaml
tthyer: done

templates/VPC/vpc-20-private.yaml
tthyer: removed

templates/VPC/vpc-20-private.yaml
tthyer: removed

tthyer:
@zaro0508 The test succeeded after adding InboundHTTPPublicNetworkAclEntry.

templates/VPC/vpc-20-private.yaml
zaro0508: don't you want to doc the usable IP range instead of the full range?  https://www.davidc.net/sites/default/subnets/subnets.html?network=192.168.0.0&mask=20&division=11.351

zaro0508 (approved):

templates/VPC/vpc-20-private.yaml
tthyer: Sure, I can make that change

PR #294, opened 2021-04-26
cloudwatch log group template

zaro0508:
Create a template to deploy a cloudwatch log group.  The output of the
log group will be passed into the cloudtrail template.  This is to
allow us to send cloudtrail logs to cloudwatch for easier viewing
and querying.

### DIFF
--- a/templates/Cloudwatch/cloudwatch-log.yaml
+++ b/templates/Cloudwatch/cloudwatch-log.yaml
@@ -1,0 +1,62 @@
+Description: Setup cloudwatch logs
+AWSTemplateFormatVersion: 2010-09-09
+Parameters:
+  RetentionInDays:
+    Type: Number
+    Description: >
+      The number of days to retain the log events in the specified log group. Possible values are:
+      1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1827, and 3653.
+    Default: 90
+    MinValue: 1
+    MaxValue: 3653
+Resources:
+  LogGroup:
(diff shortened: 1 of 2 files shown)


tthyer (approved):
LGTM! thanks!

templates/Cloudwatch/cloudwatch-log.yaml
```diff
@@ -0,0 +1,62 @@
...
+AWSTemplateFormatVersion: 2010-09-09
+Parameters:
+  RetentionInDays:
+    Type: Number
+    Description: >-
+      The number of days to retain the log events in the specified log group. Possible values are:
+      1, 3, 5, 7, 14, 30, 60, 90, 120, 150, 180, 365, 400, 545, 731, 1827, and 3653.
+    Default: 90
+    MinValue: 1
+    MaxValue: 3653
```
ahayden, line 11: Let's use this value

ahayden (approved):
Looks good

xschildw (approved):

templates/Cloudwatch/cloudwatch-log.yaml
zaro0508: do you mean use 3653 as the default?  this is a parameter which means you can override it when you deploy this template.  i think default of 90 is better than max.  you can set to max when you deploy.

ahayden: Agreed

ahayden (approved):
Looks good. Your point on not changing the default parameter makes sense. We just need to remember for when we deploy something out of band from this this repo's control

PR #297, opened 2021-05-05
parameterize LogGroupName

zaro0508:
Allow users to pass in a custom log group name.


xschildw (approved):

ahayden (approved):
Seems ok... is this because of conflicting stack names or something?

zaro0508:
possibily @ahayden  but more that users should be able to name the log group when they create one.

PR #298, opened 2021-05-06
[IT-1101] Cross account AWS cloudtrail and config access

zaro0508:
Some of our accounts, like org-sagebase-strides is not part of the Sage
organziations account therefore we need to manage them as external
entities. These templates will allow external accounts to send
cloudtrail logs to the master bucket.


xschildw (approved):
New template engine?

zaro0508:
yes, that's correct @xschildw

PR #301, opened 2021-05-10
Cross account access

zaro0508:
Add a template to allow a user in another AWS account IAM access to
resources in the current account that this template is deployed to.


ahayden (approved):
I think this is the minimal process/permissions for automating cross-account access

PR #303, opened 2021-05-13
update SSO template to allow setting inline policy

zaro0508:
AWS SSO parmeter sets does not support customer managed policies[1]
therefore we need to update the template to allow us to pass in
a inline policy.  This will allow us to define roles with custom
policies.

[1] https://forums.aws.amazon.com/thread.jspa?threadID=282793

### DIFF
--- a/templates/SSO/aws-sso.yaml
+++ b/templates/SSO/aws-sso.yaml
@@ -18,9 +18,17 @@ Parameters:
     Type: String
     Default: AdministratorAccess
 
+  # Must provide either a list of managedPolicies or an inlinePolicy.
+  # Can also provide both a list of managedPolicies and an inlinePolicy.
   managedPolicies:
     Type: CommaDelimitedList
-    Default: arn:aws:iam::aws:policy/AdministratorAccess
+    Default: ''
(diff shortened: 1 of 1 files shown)


ahayden (changes requested):
It would be surprising to be limited only to inline policies and AWS-managed policies, but maybe an inline policy in this is equivalent to using a customer managed policy, since it is centrally managed? Is possible that that managed policy just doesn't exist in both accounts with a name/reference that can be included in the template?

This doc implies that customer managed policies should work: https://docs.aws.amazon.com/singlesignon/latest/userguide/iam-auth-access-using-id-policies.html#policyexample

zaro0508:
> This doc implies that customer managed policies should work: https://docs.aws.amazon.com/singlesignon/latest/userguide/iam-auth-access-using-id-policies.html#policyexample

Well it's wrong.  I have tested using both the AWS console and CFN and I was not able to attach a customer managed policy to a SSO permission set.

ahayden (approved):
Since these inline policies are centrally managed, I believe we still have most (all) of the benefits of a customer managed policy.

PR #304, opened 2021-05-14
[IT-1205] template for VPC peering request

zaro0508:
Add a template to create a VPC peering request role. We did this
same thing in sceptre with essentials.yaml[1], this is to decouple
the role from essentials and allows org-formation to deploy it
without all the stuff in essentials.

[1] https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/essentials.yaml#L92


ahayden (approved):
This is a direct translation from essentials... I believe the process will be: (1) member account requests vpc peer to vpc in transit account and (2) transit account accepts... Correct? Wouldn't this mean only the transit account needs this role?

xschildw (approved):

PR #305, opened 2021-05-18
Output aws cloudtrail and config bucket info

zaro0508:
Add cloudformation outputs for cloudtrail and config buckets so it can
be re-used elsewhere.


ahayden (approved):
All looks good

PR #309, opened 2021-06-24
clarifying how StringNotLike works

brucehoff:
In this template it's not clear if the condition is that the userid is different from **either one** of the two or different from **both** of the two.  We want the latter, and adding the explicit qualifier, `ForAllValues` makes it clear.


brucehoff:
Disregard this PR. (I am closing it.). I was unclear what `ForAllValues` means.

zaro0508 (approved):

PR #312, opened 2021-07-02
IT-1295

brucehoff:
This PR modifies the S3 bucket template to meet the requirements listed in IT-1295, i.e.:
- Can upload from Synapse;
- Can upload from the S3 console, when logged in as the bucket owner;
- Cannot upload from a different account without adding the `bucket-owner-full-control` ACL;
- CAN upload from a different account with the `bucket-owner-full-control` ACL.  Bucket owner assumes ownership;
- A different account cannot change the ACL of an object in the bucket.

I created this policy interactively in S3, converted the JSON to YAML and mapped it back into the CFN template.  It passes the CloudFormation Linter but we'll have to check it against the UAT bucket to ensure correctness.

There are six statements in the policy:
SynapseBucketAccess - gives Synapse access to the bucket
SynapseObjectAccess - gives Synapse access to objects in the bucket (R/O or R/W)
BucketAccess - gives grantees access to the bucket
ReadObjectAccess - give grantees read access to objects
InternalPutObjectAccess - gives bucket account grantees the ability to upload objects
ExternalPutObjectAccess - gives cross-account grantees the ability to upload objects

Note that the template has a boolean 'AllowWrite'. The *PutObjectAccess statements are conditioned on this boolean.

### DIFF
--- a/templates/s3-bucket-v2.j2
+++ b/templates/s3-bucket-v2.j2
@@ -188,6 +188,7 @@ Resources:
               StringNotLike:
                 aws:userid:
                   - arn:aws:iam::325565585839:root
+                aws:userid:
                   - !Ref AWS::AccountId
               StringNotEquals:
                 s3:x-amz-acl: bucket-owner-full-control
--- a/templates/s3-bucket-v2.j2
+++ b/templates/s3-bucket-v2.j2
@@ -170,6 +170,7 @@ Resources:
         Version: "2012-10-17"
         Statement:
           - Sid: SynapseBucketAccess
+            # gives Synapse access to the bucket
             Action:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
@@ -178,6 +179,7 @@ Resources:
             Principal:
               AWS: "325565585839"
           - Sid: SynapseObjectAccess
+            # gives Synapse access to objects in the bucket (R/O or R/W, depending on AllowWrite)
             Action:
               - !If [AllowWrite, "s3:*Object*", "s3:GetObject*"]
               - "s3:*MultipartUpload*"
@@ -186,6 +188,7 @@ Resources:
             Principal:
               AWS: "325565585839"
           - Sid: BucketAccess
+            # gives grantees access to the bucket
             Effect: Allow
             Principal:
               AWS: !Ref GrantAccess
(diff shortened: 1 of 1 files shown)


templates/s3-bucket-v2.j2
```diff
@@ -169,37 +169,68 @@ Resources:
...
+              Effect: Allow
+              Principal:
+                AWS: !Ref GrantAccess
+              Action:
+                - "s3:PutObject"
+                - "s3:PutObjectAcl"
+              Resource: !If [EnableEncryption, !Sub "${SynapseEncryptedExternalBucket.Arn}/*", !Sub "${SynapseExternalBucket.Arn}/*"]
+              Condition:
+                StringEquals:
+                  s3:x-amz-acl: bucket-owner-full-control
```
zaro0508, line 233: these `if` conditions look wrong. The docs provide example as `!If [condition_name, value_if_true, value_if_false]` which implies that valid inputs to `If` condition are 3 strings, not lists.  Also you'll need to provide a 3rd parameter to the `If` construct, in this case i'm thinking you'll want a `AWS::NoValue`, https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/pseudo-parameter-reference.html

zaro0508 (changes requested):
The policy looks pretty insane now.  Would it make sense to add comments to this template describing what the policy is doing?

brucehoff:
> The policy looks pretty insane now. Would it make sense to add comments to this template describing what the policy is doing?

Done!

templates/s3-bucket-v2.j2
brucehoff: > valid inputs to If condition are 3 strings, not lists

Here's an an example that uses lists:
https://stackoverflow.com/questions/56970457/how-to-use-fnif-with-array-values-in-cloud-formation-templates

There's an example on this page too:
https://github.com/aws/serverless-application-model/issues/1218

There's an example on this page that uses a nested IF in the else clause!
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/conditions-sample-templates.html

zaro0508: ok, i'm convinced that it can take lists however i still think it's wrong.  The 1st example looks the most correct so i'm thinking it should be something like this..

```
  - !If
    - AllowWrite
    - - Sid: ExternalPutObjectAccess
         Effect: Allow
         Principal:
           AWS: !Ref GrantAccess
         Action:
           - "s3:PutObject"
           - "s3:PutObjectAcl"
         Resource: !If [EnableEncryption, !Sub "${SynapseEncryptedExternalBucket.Arn}/*", !Sub "${SynapseExternalBucket.Arn}/*"]
         Condition:
           StringEquals:
             s3:x-amz-acl: bucket-owner-full-control
    - - !Ref AWS::NoValue
```

zaro0508 (approved):

PR #315, opened 2021-07-07
Add missing subnet exports to vpc-20-private.yaml template

tthyer:
Several subnets were left off the exports from the templates/VPC/vpc-20-private.yaml template. Add these.

### DIFF
--- a/templates/VPC/vpc-20-private.yaml
+++ b/templates/VPC/vpc-20-private.yaml
@@ -285,6 +285,12 @@ Outputs:
     Export:
       Name:
         !Join ['-', [!Ref 'AWS::Region', !Sub '${AWS::StackName}', 'VpcCidr']]
+  PublicSubnet:
+    Description: "SubnetId of the public subnet"
+    Value: !Ref PrivateSubnet
+    Export:
+      Name:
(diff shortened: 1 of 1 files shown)


templates/VPC/vpc-20-private.yaml
```diff
@@ -285,6 +285,12 @@ Outputs:
     Export:
       Name:
         !Join ['-', [!Ref 'AWS::Region', !Sub '${AWS::StackName}', 'VpcCidr']]
+  PublicSubnet:
+    Description: "SubnetId of the public subnet"
+    Value: !Ref PrivateSubnet
```
xschildw, line 290: Should this not be '!Ref PublicSubnet' ?

xschildw (changes requested):
The output for PublicSubnet references the PrivateSubnet resource.

templates/VPC/vpc-20-private.yaml
tthyer: Oops, yes -- fixed in 8e3ef04

xschildw (approved):

tthyer:
Thanks @xschildw!

PR #324, opened 2021-10-01
Add S3 VPC gateway endpoints to all VPC templates

BrunoGrandePhD:
I'm propagating the change I made in Sage-Bionetworks-Workflows/nextflow-infra#67 to all VPC templates in this repository. The rationale is that it's free to run a VPC gateway endpoint, and we use S3 a lot at Sage. Hence, this can act as a preventative measure to avoid needless NAT gateway and inter-AZ traffic, which is costly. 

The resource and output added to each VPC template was taken from [`templates/gateway-vpc-endpoint.yaml`](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/gateway-vpc-endpoint.yaml). The only change that I made was adding the VPC endpoint to the public route table in addition to the private route table. This will likely have no practical change since most S3 traffic is expected to originate from private subnets, but it also can't hurt to make sure all S3 traffic from a given VPC is routed through the endpoint instead of NAT gateways. 

We've already implemented this change for [Sage-Bionetworks-Workflows/nextflow-infra](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra), so we don't need you to mint a new release once this is merged. 

depends on https://github.com/Sage-Bionetworks/iAtlas-infra/pull/28

### DIFF
--- a/templates/VPC/vpc-20-private.yaml
+++ b/templates/VPC/vpc-20-private.yaml
@@ -28,8 +28,8 @@ Parameters:
   IncludeS3GatewayEndpoint:
     Type: String
     Description: >
-      true (default) to deploy an S3 VPC gateway endpoint bucket, 
-      false to skip endpoint (e.g., one already being deployed)
+      true (default) to deploy an S3 VPC gateway endpoint,
+      false to skip endpoint (e.g., one has already been deployed)
     AllowedValues:
       - true
       - false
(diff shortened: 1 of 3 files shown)


zaro0508 (commented):
I'm wondering if this will fail if the VPC is already setup with an S3 endpoint from a different template?

PR #326, opened 2021-11-18
Allow S3 read-only actions across regions

BrunoGrandePhD:
I don't have context on what these SCPs are trying to achieve, but I want to have a discussion on whether it makes sense to allow read-only S3 actions. 

My motivating use case is running the nf-core test pipelines to ensure that Nextflow Tower works as expected. The test data for these pipelines resides in Europe (`eu-west-1`, I believe). At the moment, I cannot run these tests without first mirroring the test data into one of our buckets. I **strongly** prefer not duplicating data (for cost and convenience), so I'm hoping we can exclude read-only S3 actions from these SCPs. 

I'm also open to other ideas or specifically allowing these actions via SCP on the two Nextflow accounts. However, I think this change will be beneficial across the organization.


zaro0508 (approved):
Looks fine to me

BrunoGrandePhD:
Deploying this for the Nextflow accounts this week would help with some testing that I'm trying to do. Thanks!

PR #327, opened 2021-12-14
[SC-373] Add file for Sage program codes

zaro0508:
Add a file that will be the global reference for Sage cost
center tags in AWS.  We use json format to make it easy
to reference from other json/yaml files.


brucehoff (approved):
Please have David Lahti review.  (He doesn't use Github so I sent him a link by email.)

BrunoGrandePhD (approved):
🚀

templates/tags/SageFinancialProgramCodes.json
```diff
@@ -0,0 +1,156 @@
+[
```
zaro0508, line 1: forgot to put "Indirects" on this list.  is there a number for this?

zaro0508 (commented):
the codes in this PR were taken from https://sagebionetworks.jira.com/wiki/spaces/WF/pages/2553544733/Finance+Program+Codes

templates/tags/SageFinancialProgramCodes.json
BrunoGrandePhD: No, it's not tied to a program code. It can be added here, but I wonder if it should be filtered out for the SC. It's mostly intended for infra costs that aren't tied to a particular program (like hosting the Tower web app). 

@brucehoff: What do you think?

brucehoff: Good point. I'd forgotten about this. Let's add `NO PROGRAM | 00000` to the list. (I'd like to avoid the word "Indirect" which relates to how Finance uses the information as opposed to how resources are categorized.)

brucehoff:
@zaro0508 Please update the PR with the up-to-date program codes sent by Nikki Skillern.

templates/tags/SageFinancialProgramCodes.json
zaro0508: done

BrunoGrandePhD (approved):

PR #329, opened 2021-12-17
[SC-373] Remove invalid characters

zaro0508:
AWS has tag restriction[1] which does not allow tags with the following characters:
|, (, &, )  This PR either replaces or exchanges those characters with characters that
allowed.

[1] https://docs.aws.amazon.com/AWSEC2/latest/UserGuide/Using_Tags.html#tag-restrictions


brucehoff:
@BrunoGrandePhD , I don't recall you mentioning this.  Did you run into the same limitation?

templates/tags/SageFinancialProgramCodes.json
```diff
@@ -38,7 +38,7 @@
   "Mobile Toolbox Tech Core_112503",
   "Model AD-IU_116000",
   "NF - DoD w/UCF_122100",
-  "NF Analysis & Dissemination UCSF_121900",
+  "NF Analysis + Dissemination UCSF_121900",
```
brucehoff, line 41: I realize this is picky but can we change NO PROGRAM_00000 to NO PROGRAM_000000, i.e. since all other codes are six digits, let's use six digits here too.

brucehoff (changes requested):

templates/tags/SageFinancialProgramCodes.json
zaro0508: done

brucehoff (approved):

BrunoGrandePhD:
I just checked, and I just noticed that the new `CostCenter` tags didn't actually get deployed. **[Here](https://github.com/Sage-Bionetworks-Workflows/nextflow-infra/blob/main/config/projects/imcore-project.yaml#L23)'s** where the tag is set in the Sceptre config. I would have expected the deployment to fail. I'll update our tags accordingly. 

That said, I have a strong preference towards a slash delimiter over the underscore. 

image

BrunoGrandePhD (approved):
🚀

PR #335, opened 2022-02-05
Replace and update template to deploy windows EC2

zaro0508:
This is to replace the template to provision windows instances.

* Use PropagateTagsToVolumeOnCreation[1] property to progpogate
  tags applied to the EC2 onto the attached stroage volume. We
  no longer need custom code to propogate tags.
* Remove tags as parameter to allow passing in tags from
  cloudformation
* Move template to EC2 folder
* Add JcUserId parameter to give Jumpcloud user access to instance

[1] https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-propagatetagstovolumeoncreation

### DIFF
--- a/templates/EC2/jc-ec2-win.j2
+++ b/templates/EC2/jc-ec2-win.j2
@@ -58,9 +58,9 @@ Parameters:
     Type: String
     Default: 'yes'
   AMIId:
-    Description: Deploy with the latest released AWS AMI
-    Type: 'AWS::SSM::Parameter::Value<AWS::EC2::Image::Id>'
-    Default: '/aws/service/ami-windows-latest/Windows_Server-2022-English-Full-Base'
+    Description: The ID of the AMI to deploy
+    Type: 'AWS::EC2::Image::Id'
+    ConstraintDescription: Valid AMIs are at https://ami.sageit.org/
   VolumeSize:
     Description: The EC2 volume size (in GB)
     Type: Number
--- a/templates/EC2/jc-ec2-win.j2
+++ b/templates/EC2/jc-ec2-win.j2
@@ -1,14 +1,19 @@
 AWSTemplateFormatVersion: 2010-09-09
 Description: >-
-  Provision EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
+  Provision Windows EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
 Parameters:
+  JcUserId:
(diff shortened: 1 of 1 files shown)


templates/EC2/jc-ec2-win.j2
```diff
@@ -217,34 +199,13 @@ Resources:
...
-                  - ' -Project '
-                  - !Ref Project
-                  - ' -OwnerEmail '
-                  - !Ref OwnerEmail
-                  - ' > C:\scripts\tag-instance-volume.log'
     Properties:
       ImageId: !Ref AMIId
       InstanceType: !Ref InstanceType
       KeyName: !If [HasKeyName, !Ref KeyName, !Ref 'AWS::NoValue']
+      PropagateTagsToVolumeOnCreation: true
```
brucehoff, line 208: When I follow the [link in the PR description](https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-ec2-instance.html#cfn-ec2-instance-propagatetagstovolumeoncreation) I read, 

> PropagateTagsToVolumeOnCreation
Not currently supported by AWS CloudFormation.

Have you verified that this works?

brucehoff (changes requested):

templates/EC2/jc-ec2-win.j2
zaro0508: yes, i have verified that it works.  The doc is not up to date.

templates/EC2/jc-ec2-win.j2
```diff
@@ -1,14 +1,19 @@
 AWSTemplateFormatVersion: 2010-09-09
 Description: >-
-  Provision EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
+  Provision Windows EC2 instance, connect to Jumpcloud, and associate with a jumpcloud systems group.
```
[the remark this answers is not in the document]
zaro0508: These templates are for the case where we want to setup an EC2 in scicomp (via the scicomp provisioner) and not the SC.   The SC is for 90% of the use cases,  while scicomp is for the other 10%.   We have not done the implementation to make access the same in scicomp as SC.

Here is an example of how this template would be used.. https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/config/prod/EC2-gene-co-expression.yaml#L3

brucehoff (approved):

PR #344, opened 2022-04-20
[IT-1793] Include PasswordReusePrevention

ahayden:
[ X ] Add missing passwd reuse rule
[ X ] Passes pre-commit

### DIFF
--- a/templates/ORG/password-policy.yaml
+++ b/templates/ORG/password-policy.yaml
@@ -6,7 +6,9 @@ Resources:
     Properties:
       MinimumPasswordLength: 18
       RequireLowercaseCharacters: true
+      RequireUppercaseCharacters: true
       RequireNumbers: true
       RequireSymbols: true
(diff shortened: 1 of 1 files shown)


templates/ORG/password-policy.yaml
```diff
@@ -11,4 +11,5 @@ Resources:
       RequireSymbols: true
       AllowUsersToChangePassword: false
       HardExpiry: false
+      PasswordReusePrevention: 24
```
zaro0508, line 14: that's pretty extreme since we already require MFA.  Just wondering, is there a compliance requirement for us to be this extreme?

zaro0508 (approved):

templates/ORG/password-policy.yaml
ahayden: Yes, please see IT-1793. This should impact no one since the design is to disallow IAM users

PR #348, opened 2022-07-15
[IT-2052] Replace `jq` filter with api filter

jesusaurus:
There was a recent failure while deploying the 'jharker-test' stack in
the 'sandbox-provisioner' project during the '02_associate_jc_systems'
command. The resulting virtual machine did not have the required
association and was not accessible using ssh keys from jumpcloud.

The underlying user ID could not be found because it does not appear in the
default list of system users, but can be found when searching based on
email address.
    
Replace the client-side `jq` filter with a server-side api filter.

### DIFF
--- a/templates/EC2/jc-ec2-linux.j2
+++ b/templates/EC2/jc-ec2-linux.j2
@@ -212,12 +212,13 @@ Resources:
               command: !Join
                 - ''
                 - - "JC_SYSTEM_ID=$(sudo cat /opt/jc/jcagent.conf|jq -r '.systemKey'); "
-                  - "JC_USER_ID=$(/usr/bin/curl --silent --show-error -X GET https://console.jumpcloud.com/api/systemusers "
-                  - "-H 'Accept: application/json' -H 'Content-Type: application/json' -H 'x-api-key: "
-                  - !Ref JcServiceApiKey
-                  - "' | jq -r '.results | .[] | select(.email==\""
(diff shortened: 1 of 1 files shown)


zaro0508:
can you please describe what you suspect to be the problem?

jesusaurus:
I've re-worked this PR into a proposed solution after doing some debugging.

zaro0508:
> The underlying user ID could not be found because it does not appear in the
default list of system users

I'm wondering why your user isn't appearing on the list of system users.  Every account in JC should be on that list.

jesusaurus:
Yeah, I'm also wondering why I'm not in that list (and who else isn't).  Looking at what the `systemusers` api returns for our two users, some notable differences are that your user has `"allow_public_key": true` while my user doesn't have that key at all, and then my user shows `"public_key": ""` where your user has full public key listed. But I don't

jesusaurus:
I opened an issue with JumpCloud and heard back that the API has a hard limit of 100 users per call. They suggest making multiple calls with a `skip` parameter to collect users 100 at a time, but I prefer my original approach of using the `filter` parameter rather than using `jq` locally.

zaro0508 (commented):
I agree with this approach and it LGTM.  Was wondering whether you've run sceptre using this new `jc-ec2-linux.j2` file as a test?

zaro0508 (approved):
awsome @jesusaurus !

PR #349, opened 2022-07-26
Update synapse S3 bucket policy

zaro0508:
According to the synapse custom storage location docs[1] the
read-write permissions should contain "s3:DeleteObject" permission.

Update the policy in our templates to add that missing permission.

[1] https://help.synapse.org/docs/Custom-Storage-Locations.2048327803.html#CustomStorageLocations-Read-WritePermissions

### DIFF
--- a/templates/S3/synapse-external-bucket.j2
+++ b/templates/S3/synapse-external-bucket.j2
@@ -166,7 +166,6 @@ Resources:
               Action:
                 - "s3:PutObject"
                 - "s3:PutObjectAcl"
-                - "s3:DeleteObject"
               Resource: [ !Sub "${Bucket.Arn}/*" ]
               Condition:
                 StringEquals:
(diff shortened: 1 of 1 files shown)


jesusaurus (approved):

templates/S3/synapse-external-bucket.j2
```diff
@@ -114,7 +114,7 @@ Resources:
             Principal:
               AWS: !Ref GrantAccess
             Action:
-              - !If [AllowWrite, "s3:*Object*", "s3:GetObject*"]
+              - !If [AllowWrite, "s3:*Object*"]
```
xschildw, line 117: Don't we need an else value with !if ?

zaro0508: good catch.  it's fixed, thanks.

xschildw (approved):
Could we have relied on the ReadObjectAccess SID like the other cases?

PR #351, opened 2022-08-09
[IT-2075] Allow bucket collaboration

zaro0508:
Occasionally we want to allow our external collaborators
to use the AWS CLI to upload files to buckets that we provision.
The bucket template to create an external synapse bucket does
not support that use case currently.  We need to update the
permissions to allow for that use case.

The fix is to change from BucketOwnerPreferred to BucketOwnerEnfored[1]
to disable ACLs and set the bucket owner to autmatically have full control over
every object in the bucket.

[1] https://docs.aws.amazon.com/AmazonS3/latest/userguide/about-object-ownership.html

### DIFF
--- a/templates/S3/synapse-external-bucket.j2
+++ b/templates/S3/synapse-external-bucket.j2
@@ -90,7 +90,8 @@ Resources:
             MaxAge: 3000
       OwnershipControls:
         Rules:
-          - ObjectOwnership: BucketOwnerPreferred
+          - ObjectOwnership: BucketOwnerEnforced
+
       LifecycleConfiguration:
         Rules:
         - Id: DataLifecycleRule
@@ -143,7 +144,7 @@ Resources:
           - !If
             - AllowWrite
             - Sid: InternalPutObjectAccess
-              # gives scicomp bucket-account grantees the ability to upload and delete objects
+              # gives bucket-account grantees the ability to upload and delete objects
               Effect: Allow
               Principal:
                 AWS: !Ref GrantAccess
@@ -153,27 +154,6 @@ Resources:
                 - "s3:DeleteObject"
                 - "s3:*MultipartUpload*"
               Resource: [ !Sub "${Bucket.Arn}/*" ]
-              Condition:
-                StringEquals:
-                  "aws:PrincipalAccount": "055273631518"
-            - !Ref AWS::NoValue
-          - !If
-            - AllowWrite
-            - Sid: ExternalPutObjectAccess
-              # gives cross-account grantees the ability to upload objects
-              Effect: Allow
-              Principal:
-                AWS: !Ref GrantAccess
-              Action:
-                - "s3:PutObject"
-                - "s3:PutObjectAcl"
-                # "s3:DeleteObject" returns Conditions do not apply to combination of actions and resources in statement
-                # "s3:*MultipartUpload*" returns  Policy has invalid action
-                # "s3:DeleteObject" returns Conditions do not apply to combination of actions and resources in statement
-              Resource: [ !Sub "${Bucket.Arn}/*" ]
-              Condition:
-                StringEquals:
-                  s3:x-amz-acl: bucket-owner-full-control
             - !Ref AWS::NoValue
 
   # Add owner file to the synapse bucket, requires the cloudformation S3 objects macro
--- a/templates/S3/synapse-external-bucket.j2
+++ b/templates/S3/synapse-external-bucket.j2
@@ -128,6 +128,7 @@ Resources:
             Action:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
+              - "s3:ListBucketMultipartUploads"
             Resource: [ !GetAtt Bucket.Arn ]
           - Sid: ReadObjectAccess
             # give grantees read access to objects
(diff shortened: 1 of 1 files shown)


templates/S3/synapse-external-bucket.j2
```diff
@@ -166,6 +167,9 @@ Resources:
               Action:
                 - "s3:PutObject"
                 - "s3:PutObjectAcl"
+                # "s3:DeleteObject" returns Conditions do not apply to combination of actions and resources in statement
+                # "s3:*MultipartUpload*" returns  Policy has invalid action
+                # "s3:DeleteObject" returns Conditions do not apply to combination of actions and resources in statement
```
zaro0508, line 172: @brucehoff This is related to IT-1295.  Would you be able to provide a suggestion on how to allow cross-account grantees the ability to upload objects?

templates/S3/synapse-external-bucket.j2
```diff
@@ -128,6 +129,7 @@ Resources:
             Action:
               - "s3:ListBucket*"
               - "s3:GetBucketLocation"
+              - "s3:ListBucketMultipartUploads"
```
brucehoff, line 132: This is redundant.  If you grant ListBucket* then you also grant ListBucketMultipartUploads

templates/S3/synapse-external-bucket.j2
```diff
@@ -137,7 +139,6 @@ Resources:
             Action:
               - "s3:GetObject"
               - "s3:GetObjectAcl"
-              - "s3:AbortMultipartUpload"
```
brucehoff, line 140: What's the rationale for removing this?

templates/S3/synapse-external-bucket.j2
```diff
@@ -151,25 +152,8 @@ Resources:
                 - "s3:PutObject"
                 - "s3:PutObjectAcl"
                 - "s3:DeleteObject"
+                - "s3:*MultipartUpload*"
```
brucehoff, line 155: Clients can already upload files with the the existing permissions. Why add `*MultipartUpload*`?


If you are combining the two statements, then perhaps rename `InternalPutObjectAccess` to simply be`PutObjectAccess`.

templates/S3/synapse-external-bucket.j2
```diff
@@ -90,7 +90,8 @@ Resources:
             MaxAge: 3000
       OwnershipControls:
         Rules:
-          - ObjectOwnership: BucketOwnerPreferred
+          - ObjectOwnership: BucketOwnerEnforced
```
brucehoff, line 93: I like this!

brucehoff (changes requested):

templates/S3/synapse-external-bucket.j2
zaro0508: > Clients can already upload files with the the existing permissions. Why add *MultipartUpload*? 

Are you sure about this?  My understanding is PUT is a single operation upload while [multipart](https://docs.aws.amazon.com/AmazonS3/latest/userguide/upload-objects.html) is a special upload designed for large files.  There are specific [AWS CLI commands for multipart uploading of files](https://aws.amazon.com/premiumsupport/knowledge-center/s3-multipart-upload-cli/).

templates/S3/synapse-external-bucket.j2
brucehoff: Yes, you are right.

templates/S3/synapse-external-bucket.j2
zaro0508: @brucehoff Maybe i'm missing something but I don't see how that document clarifies whether we need `s3:*MultipartUpload*` permission here.  Maybe you can highlight where you feel it's giving us a clue?

brucehoff: @zaro0508 Please read this section: https://docs.aws.amazon.com/AmazonS3/latest/userguide/mpuoverview.html#mpuAndPermissions
which says that `s3:PutObject` lets you Create Multipart Upload, Initiate Multipart Upload, Upload Part, Complete Multipart Upload, ...

zaro0508: OK,  thanks.  However looking at this again regardless of whether `s3:PutObject` allows uploading files we still want to give `s3:*MultipartUpload*` permission to users so they can list and abort their uploads.

templates/S3/synapse-external-bucket.j2
brucehoff, line 132: This is redundant. If you grant ListBucket* then you also grant ListBucketMultipartUploads

templates/S3/synapse-external-bucket.j2
brucehoff, line 155: If you are combining the two statements, then perhaps rename InternalPutObjectAccess to simply bePutObjectAccess.

brucehoff (changes requested):

templates/S3/synapse-external-bucket.j2
zaro0508: fixed

templates/S3/synapse-external-bucket.j2
zaro0508: fixed

brucehoff (approved):

PR #355, opened 2022-08-24
IT-2000: Make bastian sg option with default on

xschildw:
The security group is getting flagged by SecurityHub EC2.19. The proposal is to make it optional for VPCs where it's not needed (default is to have it for backward compatibility).

### DIFF
--- a/templates/vpc.yaml
+++ b/templates/vpc.yaml
@@ -503,7 +503,7 @@ Resources:
   # Allow access to bastian hosts
   BastianSecurityGroup:
     Type: 'AWS::EC2::SecurityGroup'
-    Condition: EnableBastionSecurityGroup
+    Condition: EnableBastianSecurityGroup
     Properties:
       GroupDescription: Security Group for bastians
       VpcId:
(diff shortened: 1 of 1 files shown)


templates/vpc.yaml
```diff
@@ -51,8 +51,18 @@ Parameters:
...
     Default: true
+  IncludeBastianSecurityGroup:
+    Type: String
+    Description: >
+      true (default) to deploy a bastian security group
+      false to skip the bastian security group
+    AllowedValues:
+      - true
+      - false
+    Default: true
```
zaro0508, line 62: I don't see how this fixes issue IT-2000.  This default is telling CFN to create this SG unless it's explicitly set to `false`.  Do you intend to set this to false on every VPC deployment?

zaro0508 (changes requested):

templates/vpc.yaml
xschildw: I thought about setting false as default but that would have removed the sg everywhere (my concern was that some account that uses it would suddenly lose it). So I put true to keep things as is, then we explicitly specify true where we needed, then set default as false to remove (too contrived?).

zaro0508 (approved):

PR #356, opened 2022-08-29
IT-2097: Change default CIDR for VPN

xschildw:
This PR changes the default VPN CIDR from 10.1.0.0/16 to 10.50.0.0/16.
Passed pre-commit.

### DIFF
--- a/templates/vpc.yaml
+++ b/templates/vpc.yaml
@@ -479,11 +479,6 @@ Resources:
           ToPort: -1
           IpProtocol: "-1"
           Description: "Allow all VPN traffic"
-        - CidrIp: "10.50.0.0/16"    # AWS TGW Hub
-          FromPort: -1
-          ToPort: -1
-          IpProtocol: "-1"
-          Description: "Allow AWS TGW Hub traffic"
       # CF does not support removing all rules, workaround is to add a pointless rule
       SecurityGroupEgress:
         - CidrIp: "0.0.0.0/0"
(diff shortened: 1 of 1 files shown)


templates/vpc.yaml
```diff
@@ -19,9 +19,9 @@ Parameters:
     ConstraintDescription: List of Availability Zones in a region, such as us-east-1a, us-east-1b, us-east-1c
     Default: "us-east-1a, us-east-1b, us-east-1c"
   VpnCidr:
-    Description: CIDR of the (sophos-utm) VPN
+    Description: CIDR of the TGW VPN
     Type: String
-    Default: "10.1.0.0/16"
+    Default: "10.50.0.0/16"
```
zaro0508, line 24: there's a duplicate static entry on line 533.  should we remove that as well?

zaro0508 (changes requested):

templates/vpc.yaml
xschildw: You mean the  !Ref? That should be fine shouldn't it (just change from 10.1.* to 10.50.*)?

zaro0508: No I meant this section..
```
      SecurityGroupIngress:
        - CidrIp: !Ref VpnCidr
          FromPort: -1
          ToPort: -1
          IpProtocol: "-1"
          Description: "Allow all VPN traffic"
        - CidrIp: "10.50.0.0/16"    # AWS TGW Hub
          FromPort: -1
          ToPort: -1
          IpProtocol: "-1"
          Description: "Allow AWS TGW Hub traffic"
```

now that you've changed `!Ref VpnCidr` to a default of `10.50.0.0/16` isn't it going to be the same as the `AWS TGW Hub`?  Therefore shouldn't we just git rid of the static (AWS TGW Hub) CidrIp config?

zaro0508 (changes requested):

templates/vpc.yaml
xschildw: Indeed...

zaro0508 (approved):

PR #358, opened 2022-10-06
Setup cost center for RECOVER project

zaro0508:
Add a cost center code for the RECOVER project.

### DIFF
--- a/templates/tags/SageFinancialProgramCodesOther.json
+++ b/templates/tags/SageFinancialProgramCodesOther.json
@@ -51,6 +51,7 @@
   "PECDCC-UMass / 106000",
   "PKU App-BioMarin / 613400",
   "Psorcast Pfizer / 613600",
+  "RECOVER / 122500",
   "Risk Variants - UW / 120900",
   "Scientific Program Management / 990200",
   "Synodos-CTF / 304000",


templates/tags/SageFinancialProgramCodesOther.json
```diff
@@ -51,6 +51,7 @@
   "PECDCC-UMass / 106000",
   "PKU App-BioMarin / 613400",
   "Psorcast Pfizer / 613600",
+  "RECOVER / 122500",
```
jesusaurus, line 54: We already have `MGH RECOVER / 122500` above on line 36, which matches the name in our financial system

jesusaurus (changes requested):

templates/tags/SageFinancialProgramCodesOther.json
zaro0508: welp, good thing we review stuff here. thanks @jesusaurus!  

@itismeghasyam i'm guessing that you can use the existing one and I'll just close out this PR.

itismeghasyam: I will change it to 'MGH RECOVER /  122500'. Thanks!

PR #359, opened 2022-10-12
Update s3 policy to allow removing delete markers

zaro0508:
Delete markers can get created on failed file uploads which can block users from re-uploading the same file. The user must be able to remove the delete marker before re-uploading the file. This will allow s3:DeleteObjectVersion[1] permission which will  users to remove delete markers.

[1] https://aws.amazon.com/premiumsupport/knowledge-center/s3-undelete-configuration

### DIFF
--- a/templates/S3/synapse-external-bucket.j2
+++ b/templates/S3/synapse-external-bucket.j2
@@ -150,7 +150,7 @@ Resources:
               Action:
                 - "s3:PutObject"
                 - "s3:PutObjectAcl"
-                - "s3:DeleteObject"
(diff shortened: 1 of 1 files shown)


xschildw (approved):

allaway:
@zaro0508 , I am not sure that "which can block users from re-uploading the same file." is correct. We thought that might be the case yesterday, but the uploader is still having issues after deleting these markers. I am not sure whether that impacts this PR though -maybe it would be good to still allow users to remove delete markers.

PR #360, opened 2022-10-18
[IT-2268] Add support for CustomerManagedPolicyReference

jesusaurus:
In order to use a custom managed policy for granting access to the cost explorer, we need to add support for at least one customer-managed policy to the managed-policy template. Adding support for multiple customer-managed policies in a single permission set will be more difficult due to the data structure of the parameter.

Updates to cfn-lint are required to avoid an E3002, and the updates also introduced new E2531 and E1001 occurrences - the former is a deprecation error for python 3.6, and the latter is related to org-formation syntax.

### DIFF
--- a/templates/SSO/aws-sso.njk
+++ b/templates/SSO/aws-sso.njk
@@ -25,11 +25,6 @@ Parameters:
     Default: ''
     Description: 'A list of AWS managed policy ARNs for a SSO permission set'
 
-  customerManagedPolicyName:
-    Type: String
-    Default: ''
-    Description: 'The name of an optional customer managed policy for a SSO permission set'
-
   inlinePolicy:
     Type: String
     Default: ''
@@ -48,7 +43,6 @@ Conditions:
   includePermissionSet: !Equals [ '', !Ref permissionSetArn ]
   includeMaster: !Not [ !Equals [ '', !Ref masterAccountId ] ]
   includeInlinePolicy: !Not [ !Equals [ '', !Ref inlinePolicy ] ]
-  includeCustomerManagedPolicy: !Not [ !Equals [ '', !Ref customerManagedPolicyName ] ]
   includeManagedPolicies: !Not
     - !Equals
       - !Join ['', !Ref managedPolicies]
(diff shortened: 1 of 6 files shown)


xschildw (approved):

templates/SSO/aws-sso.yaml
```diff
@@ -25,6 +25,11 @@ Parameters:
     Default: ''
     Description: 'A list of AWS managed policy ARNs for a SSO permission set'
 
+  customerManagedPolicyName:
+    Type: String
```
zaro0508, line 29: can we make this a list, similar to `managedPolicies`, so that users can pass in a list of customer policies?

templates/SSO/aws-sso.yaml
```diff
@@ -58,6 +64,13 @@ Resources:
       Description: !Sub '${permissionSetName} access to AWS organization'
       InstanceArn: !Ref instanceArn
       ManagedPolicies: !If [ includeManagedPolicies, !Ref managedPolicies, !Ref 'AWS::NoValue' ]
+      CustomerManagedPolicyReferences:
+        !If
+          - includeCustomerManagedPolicy
+          - # If we are setting a value, we are also embedding it as a single item in a list
```
zaro0508, line 70: this, dash with comment, looks wrong.  value probably should look something like value for ManagedPolicies, no?

zaro0508 (changes requested):

templates/SSO/aws-sso.yaml
jesusaurus: I think that would require a macro to parse the list of strings and turn it into a list of objects expected by the parameter, similar to my cost categories rules generator macro, and I didn't want to put in that much effort when we only have a need for a single customer managed policy.

templates/SSO/aws-sso.yaml
jesusaurus: We have two lists here: the outer list used by the `!If` where the second element in the list is our desired output, and the output value itself, which is a list with a single element. I've confirmed with a manual sceptre deploy that this works as expected.

templates/SSO/aws-sso.yaml
zaro0508: ahh, it's because customer managed policies would be a dictionary of items.  i think the solution is to templatetize it using org-formation's [TemplatingContext](https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#templating).  Fist step is to convert this template to a nunjucks template. Then pass in a dictionary value for `customerManagedPolicyNames` with  TemplatingContext then reference the dict in the template.

jesusaurus: oh, awesome! thanks

jesusaurus: I've converted the template to nunjucks.

templates/SSO/aws-sso.njk
```diff
@@ -58,6 +58,13 @@ Resources:
       Description: !Sub '${permissionSetName} access to AWS organization'
       InstanceArn: !Ref instanceArn
       ManagedPolicies: !If [ includeManagedPolicies, !Ref managedPolicies, !Ref 'AWS::NoValue' ]
+      {% if ( customerManagedPolicies | default(false) ) %}
+      CustomerManagedPolicyReferences:
+        {% for policy in customerManagedPolicies %}
+        - Name: {{ policy['Name'] }}
```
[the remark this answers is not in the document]
jesusaurus: according to https://mozilla.github.io/nunjucks/templating.html#variables they are equivalent

zaro0508 (approved):

PR #361, opened 2022-10-26
[IT-2221] Add flag for bastian access to vpc-mini

jesusaurus:
Copy the IncludeBastianSecurityGroup parameter from the vpc template to the vpc-mini template so that we can disable access to `scheduledjobsvpc` in scipool prod.

### DIFF
--- a/templates/VPC/vpc-mini.yaml
+++ b/templates/VPC/vpc-mini.yaml
@@ -34,8 +34,18 @@ Parameters:
       - true
       - false
     Default: true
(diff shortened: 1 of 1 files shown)


xschildw (approved):

zaro0508 (approved):
@xschildw did we miss updating the VPN CIDR in [vpc-20-private.yaml](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/VPC/vpc-20-private.yaml#L22-L27) ?

PR #363, opened 2022-11-09
[IT-1665] Remove travis service user

zaro0508:
This template is old and we shouold not be using the travis user that's created from it anymore.  We should not be using CI users that are created from org formation access deployments[1]

NOTE: This may cause build failures in travis-ci because really old CI configs might still be using these service users.  The strategy is to fix up as needed.

[1] https://github.com/Sage-Bionetworks-IT/organizations-infra/tree/master/org-formation/600-access

### DIFF
--- a/templates/bootstrap.yaml
+++ b/templates/bootstrap.yaml
@@ -29,25 +29,6 @@ Parameters:
     ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
     Default: 'it@sagebase.org'
 Resources:
-  # !! IMPORTANT !! - AWS API will refuse to remove users that have attached resources.
-  # Therefore you must do the following before deleting them from this file:
-  # 1. Detach or remove the following user resources: login profile, attached
-  #    MFA device, access-keys, ssh-keys, and policies.
-  # 2. Detach the user from all groups.
-  AWSIAMTravisUserAccessKey:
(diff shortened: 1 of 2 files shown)


brucehoff (approved):

templates/bootstrap.yaml
```diff
@@ -29,25 +29,6 @@ Parameters:
     ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
     Default: 'it@sagebase.org'
 Resources:
-  # !! IMPORTANT !! - AWS API will refuse to remove users that have attached resources.
-  # Therefore you must do the following before deleting them from this file:
-  # 1. Detach or remove the following user resources: login profile, attached
-  #    MFA device, access-keys, ssh-keys, and policies.
-  # 2. Detach the user from all groups.
-  AWSIAMTravisUserAccessKey:
```
jesusaurus, line 37: Don't we use this key in this repo for pushing the templates to S3?

jesusaurus (changes requested):

templates/bootstrap.yaml
zaro0508: i believe we actually use the role this template from https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/admincentral/templates/service-accounts.yaml

jesusaurus (approved):

PR #366, opened 2022-11-15
[IT-1655] Sync latest program codes

jesusaurus:
Soon we will be able to replace these static files with calls to an api lambda, but for now update the files with the latest program codes from the MIPS system, sorting alphabetically.


templates/tags/SageFinancialProgramCodesOther.json
```diff
@@ -1,44 +1,48 @@
...
   "AMP 2.0 Diverse Cohorts AD Portal / 102300",
   "AMP AD Supp Diversity Cohorts-Columbia / 121300",
   "AMP AD Supp Diversity Cohorts-Mayo / 121400",
   "AMP Longevity Supplement / 102000",
   "ANTI PD1 CH-BMS / 507200",
+  "Agora / 112300",
+  "Allen Institute DREAM CH / 507400",
   "AoU Governance-ELSI / 122901",
   "At Home-Mass Gen Hosp / 120000",
-  "BMGF-Ki / 30144",
```
zaro0508, line 11: We need to keep some of these around even though they no longer exist in the finance app.  It's because these have been used to tag resources and we need to keep the resources around even after the code disappears.  here's an example: https://github.com/Sage-Bionetworks/scicomp-provisioner/blob/master/config/prod/gates-ki-001.yaml

zaro0508 (changes requested):

templates/tags/SageFinancialProgramCodesOther.json
brucehoff: Khai is right.  It might be safest to make additions only in this PR, omitting deletions. (If you maintain the original sort order you can see the additions and deletions a little easier in the PR.)

jesusaurus: I've re-added the inactive code, but vim's case-insensitive sorting has moved it down a couple rows

zaro0508 (approved):

PR #368, opened 2022-12-01
[IT-2043] Use pip to install aws-cfn-bootstrap

jesusaurus:
Use pip and venv instead of easy_install when installing aws-cfn-bootstrap to support more recent versions of Ubuntu.

### DIFF
--- a/templates/EC2/jc-ec2-linux.j2
+++ b/templates/EC2/jc-ec2-linux.j2
@@ -251,9 +251,9 @@ Resources:
           #!/bin/bash -xe
 {% if sceptre_user_data.Distro.lower() == 'ubuntu' %}
           apt-get update -y
-          mkdir -p /opt/aws/bin
-          wget https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz
-          python3 -m easy_install --script-dir /opt/aws/bin aws-cfn-bootstrap-py3-latest.tar.gz
+          apt-get -y install python3-pip python3-venv
+          python3 -m venv /opt/aws
+          /opt/aws/bin/pip3 install https://s3.amazonaws.com/cloudformation-examples/aws-cfn-bootstrap-py3-latest.tar.gz
 {% endif %}
           /opt/aws/bin/cfn-init -v --stack ${AWS::StackName} --resource Ec2Instance --configsets SetupCfn,SetupTools,SetupJumpcloud --region ${AWS::Region}
           /opt/aws/bin/cfn-signal -e $? --stack ${AWS::StackName} --resource Ec2Instance --region ${AWS::Region}


brucehoff (approved):

zaro0508 (changes requested):
So the [official aws docs](https://aws.amazon.com/premiumsupport/knowledge-center/install-cloudformation-scripts/) that reference [examples scripts](https://github.com/awslabs/aws-cloudformation-templates/tree/master/aws/solutions/OperatingSystems) says to use easy_install.  I'm wondering if we should stick with the official docs?  

Also there is a [reference](https://github.com/ACloudGuru/AdvancedCloudFormation/issues/2) in the jira issue that seems to indicate the easy_install is still in ubuntu 22.

jesusaurus:
@zaro0508 ubuntu 20.04 has a version of setuptools that deprecated support for easy_install by removing the executable scripts, and ubuntu 22.04 ships with a version of setuptools that does not officially support easy_install. technically, we can install `python-setuptools` to get an ancient version of setuptools that supports easy_install by using python 2.7 and pinning aws-cfn-bootstrap to a version that supports python 2.7, but that would be a pretty insecure solution.

I'm not 100% sure what easy_install does different from pip, so if we don't want to experiment with using pip for the install maybe we should wait for aws-cfn-bootstrap to officially support pip before trying to install it on ubuntu 22.04

zaro0508 (commented):
We have existing deployments of ubuntu instances and my concern is that this change will attempt to update those instances which could do 2 things that our customers may not like.  The first is that it will auto terminate and create new instances which will wipe out all the data on the instances.  The second is that this change may fail to work on the AMIs used in existing deployments.

```
➜  scicomp-provisioner $ git grep 'Distro: "ubuntu"'
config/prod/minidream-rstudio-2022-ec2.yaml:  Distro: "ubuntu"
config/prod/psorcast-ec2.yaml:  Distro: "ubuntu"
config/prod/shinyserver-pro-prod.yaml:  Distro: "ubuntu" 
```

zaro0508 (approved):
Actually we version these files which invalidates my concern.  i think this change is fine but please test it with the latest AWS ubuntu AMI before merging and tagging.

```
➜  git grep jc-ec2-linux.j2
config/prod/minidream-rstudio-2022-ec2.yaml:  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.4.0/EC2/jc-ec2-linux.j2"
config/prod/psorcast-ec2.yaml:  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.4.0/EC2/jc-ec2-linux.j2"
config/prod/shinyserver-pro-prod.yaml:  url: "https://{{stack_group_config.admincentral_cf_bucket}}.s3.amazonaws.com/aws-infra/v0.4.0/EC2/jc-ec2-linux.j2"
```

jesusaurus:
I was able to successfully test this on `ami-06878d265978313ca`

PR #370, opened 2022-12-15
[IT-2431] Require HTTPS for guard duty trusted IPs

jesusaurus:
Add bucket policy denying any traffic not over https to address CIS 2.1.2.

[1] https://aws.amazon.com/premiumsupport/knowledge-center/s3-bucket-policy-for-config-rule/


zaro0508 (approved):
how about adding the link referenced in this PRs description as a comment here?

PR #372, opened 2023-02-01
[IT-2594] Create template for setting a DNS CNAME

jesusaurus:
Create a template inspired by the s3-redirector template, but intended to redirect to a Cloudfront distribution created in another stack.

This is needed to create a friendly DNS name for the lambda-mips-api Cloudfront distribution.

### DIFF
--- a/templates/R53/cname.yaml
+++ b/templates/R53/cname.yaml
@@ -1,0 +1,31 @@
+# This template sets up a Route 53 CNAME entry
+# Partial copy of `templates/s3-redirector.yaml`
+AWSTemplateFormatVersion: 2010-09-09
+Description: Provision a Route 53 CNAME at `SourceHostName` that redirects to `TargetHostName`.
+Parameters:
+  SourceHostName:
+    Description: The name of the new CNAME record to create
+    Type: String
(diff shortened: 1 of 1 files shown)


zaro0508 (changes requested):
i think the [s3-redirector.yaml](https://github.com/Sage-Bionetworks/aws-infra/blob/master/templates/s3-redirector.yaml) template is already doing the same thing?  

it's used here https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/800-redirects/_tasks.yaml#L6

xschildw (commented):
I would remove references to 'redirect', there's no redirection (in the HTTP 30x sense) it's just different names that point to the same IP addresses. If we do a redirect, we'll end up on the target with its raw/internal name (maybe it does not matter since it's not meant to be seen in a browser).

jesusaurus:
@zaro0508 the s3-redirector template is doing too much for our lambda CNAME use-case, this template is only creating the CNAME. And it might be nice to refactor s3-redirector to leverage this template instead of creating a duplicate resource definition, but I think that should be a follow-up effort.

zaro0508 (approved):
@xschildw comment is interesting however i don't really understand what it means.  LGTM.

PR #374, opened 2023-02-08
Remove obsolete import from bootstrap

xschildw:
This PR removes the import for an output deleted in bootstrap in #363.

### DIFF
--- a/templates/essentials.yaml
+++ b/templates/essentials.yaml
@@ -74,7 +74,6 @@ Resources:
             Principal:
               AWS:
                 - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
-                - !ImportValue us-east-1-bootstrap-CfServiceRoleArn
             Action:
               - "kms:Create*"
               - "kms:Describe*"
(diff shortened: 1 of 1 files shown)


jesusaurus (approved):

templates/essentials.yaml
```diff
@@ -74,7 +74,6 @@ Resources:
             Principal:
               AWS:
                 - !Sub 'arn:aws:iam::${AWS::AccountId}:root'
-                - !ImportValue us-east-1-bootstrap-CfServiceRoleArn
```
zaro0508, line 77: i think whatever role that the CI is using for deployment still needs access otherwise the CI will not be able toupdatethis resource.  For Sage-Bionetworks-IT/organizations-infra it will be `!ImportValue us-east-1-sagebase-github-oidc-sage-bionetworks-it-ProviderRoleArn-organizationsinfra`

templates/essentials.yaml
```diff
@@ -95,7 +94,6 @@ Resources:
...
```
zaro0508, line 98: here as well.

templates/essentials.yaml
xschildw: Good catch! I thought that key was next to go...

PR #376, opened 2023-02-17
IT-2672: Migrate aws-infra deployment to GitHub actions

xschildw:
This PR moves the deployment of the aws-infra templates to S3 from Travis to GitHub workflows.
This is in support of removing the BootstrapTravisUser (IT-2635) from the admincentral account.

Depends on Sage-Bionetworks-IT/organizations-infra/pull/746 (and fix in 748)


.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,62 @@
+name: aws-deploy
+
+on:
+  push:
+    branches: [ '*' ]
+    tags: [ 'v*' ]
```
zaro0508, line 6: let's make this regex a tighter match, try something like this `v[0-9]+\.[0-9]+\.[0-9]+$`

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+        run: |
+          python -m pip install --upgrade jinja2==3.0.1
+      - name: Install jinja linter tool
+        run: git clone --single-branch https://github.com/drm/jinja2-lint.git /tmp/jinja2-lint && pushd /tmp/jinja2-lint && git checkout 75dcd5a
+      - name: Execute jinja linter
+        run: find . -type f -name "*.j2" -exec /tmp/jinja2-lint/j2lint.py '{}' +
+  deploy:
+    name: Upload to Amazon S3
+    runs-on: ubuntu-latest
+    if: ${{ github.ref == 'refs/heads/main' || startsWith(github.ref, 'refs/tags/v') }}
```
zaro0508, line 45: tighter regex for tag

zaro0508 (changes requested):

.github/workflows/deploy.yaml
```diff
@@ -0,0 +1,62 @@
...
+    - name: Assume AWS Role
+      uses: aws-actions/configure-aws-credentials@v1
+      with:
+        aws-region: us-east-1
+        role-to-assume: arn:aws:iam::745159704268:role/sagebase-github-oidc-sage-bio-ProviderRoleawsinfra-17DRC23HA177D
+        role-session-name: GitHubActions-${{ github.repository_owner }}-${{ github.event.repository.name }}-${{ github.run_id }}
+        role-duration-seconds: 120
+    - name: Copy files with the AWS CLI
+      run: |
+        aws s3 cp templates s3://bootstrap-awss3cloudformationbucket-19qromfd235z9/$REPO_NAME/$BRANCH_OR_TAG/ --recursive
```
zaro0508, line 62: `REPO_NAME` variable is defined in .travis.yml file but not in this file.  Also where does `$BRANCH_OR_TAG` variable come from?

Maybe you can change this to use one of the S3 github actions from the marketplace?

zaro0508 (changes requested):

.github/workflows/deploy.yaml
xschildw: REPO_NAME and BRANCH_OR_TAG are defined in the 'env' section. Will do on S3 GHA.

zaro0508: ahh ok, sorry i don't know why i missed that. you can just use them directly here instead of setting them in `env`.  also it might be better to switch this to a GH action on a follow on PR.

zaro0508 (approved):

.github/workflows/deploy.yaml
xschildw: Removed env vars and used GH pseudovars directly, holding in GHA for s3 cp.

zaro0508 (approved):

PR #378, opened 2023-02-23
Remove deprecated Project, Depantment, and OwnerEmail tags

jesusaurus:
These tags are not used, remove them before creating new resources from these templates.

### DIFF
--- a/templates/R53-hostedzone.yaml
+++ b/templates/R53-hostedzone.yaml
@@ -1,12 +1,6 @@
 Description: Setup a hosted zone for DNS
 AWSTemplateFormatVersion: 2010-09-09
 Parameters:
-  OwnerEmail:
-    Description: 'Email address of the owner of this resource'
-    Type: String
-    AllowedPattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
-    ConstraintDescription: 'Must be an acceptable email address syntax(i.e. joe.smith@sagebase.org)'
-    Default: 'it@sagebase.org'
   DnsDomainName:
     Description: A fully qualified domain name (i.e. sagebionetworks.org)
     Type: String
@@ -21,9 +15,6 @@ Resources:
         -
           Key: "Name"
           Value: !Ref AWS::StackName
-        -
-          Key: "OwnerEmail"
-          Value: !Ref OwnerEmail
 Outputs:
   HostedZoneId:
     Value: !Ref HostedZone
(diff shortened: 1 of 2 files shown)


xschildw (approved):

templates/R53-hostedzone.yaml
```diff
@@ -1,18 +1,21 @@
 Description: Setup a hosted zone for DNS
 AWSTemplateFormatVersion: 2010-09-09
 Parameters:
-  Department:
-    Description: The department for this resource (i.e. Computational Oncology)
-    Type: String
-  Project:
-    Description: The name of the project that this resource is used for (i.e. Resilience)
+  CostCenter:
```
zaro0508, line 4: I don't think we need this if we just pass in a CostCenter cloudformation tag on deployment which i think is the preferred approach.

zaro0508 (changes requested):

templates/R53-hostedzone.yaml
jesusaurus: I don't understand what you mean by passing in a tag on deployment, without tagging the resource itself we would have to rely on the account tag for the containing account (which is fine for now, but I would prefer to have the option to set a CostCenter tag on all resources that support tagging)

zaro0508: `R53-hostedzone.yaml` is a template.  when deploying the template (via cloudformation) you can [pass tags to cloudformation](https://docs.aws.amazon.com/cli/latest/reference/cloudformation/deploy/index.html) which will recursively apply those tags to all resource in a template.  When deploying with [org-formation update-stacks](https://github.com/org-formation/org-formation-cli/blob/master/docs/task-files.md#update-stacks) setting the tags will pass those tags to cloudformation.

jesusaurus: oh, I forgot that org-formation can propagate those tags; yeah let's rely on that instead of passing the values into every template

templates/R53-hostedzone.yaml
```diff
@@ -1,15 +1,12 @@
 Description: Setup a hosted zone for DNS
 AWSTemplateFormatVersion: 2010-09-09
 Parameters:
-  Department:
-    Description: The department for this resource (i.e. Computational Oncology)
-    Type: String
-  Project:
-    Description: The name of the project that this resource is used for (i.e. Resilience)
-    Type: String
   OwnerEmail:
```
zaro0508, line 4: OwnerEmail is also used a tag which I don't think we need either.  is there any reason why you are keeping this around?

zaro0508 (changes requested):

templates/R53-hostedzone.yaml
jesusaurus: yeah, it's not as well implemented as the CostCenter cost category (there are many more untagged resources), but there's also an effort to create an Owner cost category, and this is one of the tags that cost category is derived from

zaro0508: the question is do we need `OwnerEmail` in this template?  can we just pass it in as a cloudformation tag?

jesusaurus: I can imagine a scenario where a project team uses this template and wants to have a specific team member responsible for DNS, but combing through our current usage this doesn't seem to be the case today - and taking a step back, we would probably want the template isolated to its own cloudformation stack in that case, so I'm going to remove this as well and we can add it back if we need to.

zaro0508 (approved):

PR #379, opened 2023-03-08
IT-2672: Fix duration on credentials action

xschildw:
This PR sets the role duration to 900 (min per error msg).


zaro0508 (approved):
weird we are using 120 elsewhere without problems.  seem fine to me though.

There is a `v2` for aws-actions/configure-aws-credentials@v2 now so you'll probably want to switch to that after getting this to work. Maybe do it on a follow on PR?

PR #382, opened 2023-04-14
Remove Dept and Project params

xschildw:
This PR removes the Department and Project parameters from the template (not needed anymore).


jesusaurus (changes requested):
I would prefer to do this to all templates at once and increase the minor version with the change.

PR #383, opened 2023-04-14
[IT-2703] Remove old tags from template parameters

jesusaurus:
Remove the Department and Project tags because we don't use them, and remove OwnerEmail because it should be applied to the entire stack and propagated to all resources within it.

We should increase the minor version to 0.7.x with this change.

### DIFF
--- a/templates/vpc.yaml
+++ b/templates/vpc.yaml
@@ -24,24 +24,6 @@ Parameters:
     Default: "10.50.0.0/16"
     AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(1[6-9]|2[0-8]))$
     ConstraintDescription: CIDR block parameter must be in the form x.x.x.x/16-28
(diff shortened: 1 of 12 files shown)


xschildw (approved):

templates/vpc.yaml
```diff
@@ -24,24 +24,6 @@ Parameters:
     Default: "10.50.0.0/16"
     AllowedPattern: ^(([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])\.){3}([0-9]|[1-9][0-9]|1[0-9]{2}|2[0-4][0-9]|25[0-5])(\/(1[6-9]|2[0-8]))$
     ConstraintDescription: CIDR block parameter must be in the form x.x.x.x/16-28
-  Department:
```
zaro0508, line 27: I'm not concerned about the other ones but for some reason I've experience problems when removing tags on this template.  I think i've tried this in the past and AWS failed to update leaving the stack in a bad state.  I think we should probably try this on the dustbunny VPC in itsandbox account before making this update.

jesusaurus: @zaro0508 thanks, good to know, I'll make sure to test with that vpc once this merges, and roll back the change for this template if it doesn't work.

PR #393, opened 2023-06-16
PLFM-7886: Adding a template for an IAM Role

brucehoff:
PLFM-7886 calls for enabling AWS Quicksight.
Quicksight requires a *role* it can assume, which has access to needed data sources.
`aws-infra` has no generic template for defining a role, so this PR adds such a template.

### DIFF
--- a/templates/IAM/iam-role.yaml
+++ b/templates/IAM/iam-role.yaml
@@ -8,15 +8,15 @@ Parameters:
     Description: Policy describing trust relationships
     Default: ''
   Policies:
-    Type: String
-    Description: In-line policy
-    Default: ''
+    Type: CommaDelimitedList
+    Description: In-line policies
+    Default: []
   ManagedPolicyArns:
-    Type: String
-    Description: ARNs of managed policies
-    Default: ''
+    Type: CommaDelimitedList
+    Default: []
+    Description: A list of managed policies for the role.
   MaxSessionDuration:
-    Type: Integer
+    Type: Number
     Description: Maximum session duration in seconds
     Default: 43200
   RoleName:
--- a/templates/IAM/iam-role.yaml
+++ b/templates/IAM/iam-role.yaml
@@ -1,0 +1,46 @@
+# Define a role
+#
+AWSTemplateFormatVersion: '2010-09-09'
+Description: 'Creates a role'
+Parameters:
+  AssumeRolePolicyDocument:
+    Type: String
+    Description: Policy describing trust relationships
+    Default: ''
+  Policies:
+    Type: String
+    Description: In-line policy
+    Default: ''
+  ManagedPolicyArns:
+    Type: String
+    Description: ARNs of managed policies
+    Default: ''
+  MaxSessionDuration:
+    Type: Integer
+    Description: Maximum session duration in seconds
+    Default: 43200
+  RoleName:
+    Type: String
+    Description: Role name
+Resources:
+  Role:
+    Type: AWS::IAM::Role
+    Properties:
+      AssumeRolePolicyDocument: !Ref AssumeRolePolicyDocument
+      Policies: !Ref Policies
+      ManagedPolicyArns: !Ref ManagedPolicyArns
+      MaxSessionDuration: !Ref MaxSessionDuration
+      RoleName: !Ref RoleName
+Outputs:
+  RoleName:
+    Value: !Ref Role
+    Export:
+      Name: !Sub '${AWS::Region}-${AWS::StackName}-RoleName'
+  RoleId:
+    Value: !GetAtt Role.RoleId
+    Export:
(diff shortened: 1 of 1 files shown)


zaro0508 (changes requested):
This template is just a wrapper for `AWS::IAM::Role`.  I don't see what value it provides and  I'm not sure why you would want to use this wrapper when you can just pass in all of the property values directly to `AWS::IAM::Role`?

Also I think the `template/IAM/cross-account-access.yaml` template is probably a better version of this template.

brucehoff:
> This template is just a wrapper for `AWS::IAM::Role`. I don't see what value it provides and I'm not sure why you would want to use this wrapper when you can just pass in all of the property values directly to `AWS::IAM::Role`?
> 
> Also I think the `template/IAM/cross-account-access.yaml` template is probably a better version of this template.

I can just put the stack definition into one of those [_tasks.yaml](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/200-baseline/_tasks.yaml) files, without involving a template?  Do we use that pattern elsewhere in https://github.com/Sage-Bionetworks-IT/organizations-infra?  Would it look something like [this](https://docs.sceptre-project.org/dev/docs/get_started.html#our-first-stack-config-vpc-yaml)?

zaro0508:
> I can just put the stack definition into one of those [_tasks.yaml](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/200-baseline/_tasks.yaml) files, without involving a template? Do we use that pattern elsewhere in https://github.com/Sage-Bionetworks-IT/organizations-infra? Would it look something like [this](https://docs.sceptre-project.org/dev/docs/get_started.html#our-first-stack-config-vpc-yaml)?

Sorry i see the purpose now.  however i still don't see how it's useful to just create a role without a policy to go with it?  Nothing can assume this role without an attached policy.

brucehoff:
> > I can just put the stack definition into one of those [_tasks.yaml](https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/200-baseline/_tasks.yaml) files, without involving a template? Do we use that pattern elsewhere in https://github.com/Sage-Bionetworks-IT/organizations-infra? Would it look something like [this](https://docs.sceptre-project.org/dev/docs/get_started.html#our-first-stack-config-vpc-yaml)?
> 
> Sorry i see the purpose now. however i still don't see how it's useful to just create a role without a policy to go with it? Nothing can assume this role without an attached policy.

Sorry, I am confused. The `ManagedPolicyArns` parameter allows the user to specify the policy. (Also, the `AssumeRolePolicyDocument` parameter allows specifying the trust relationship, which says who/what can assume the role.) So the role *would* be created with a policy (and a trust relationship). Is there something else that I am missing?

templates/IAM/iam-role.yaml
```diff
@@ -0,0 +1,41 @@
+# Define a role
+#
+AWSTemplateFormatVersion: '2010-09-09'
+Description: 'Creates a role'
+Parameters:
+  AssumeRolePolicyDocument:
+    Type: String
+    Description: Policy describing trust relationships
+    Default: ''
```
zaro0508, line 9: if you make this optional then you need to handle it with a condition, otherwise if the user doesn't pass in something for this the deployment will fail.

templates/IAM/iam-role.yaml
```diff
@@ -0,0 +1,41 @@
...
+  ManagedPolicyArns:
+    Type: CommaDelimitedList
+    Default: ''
```
zaro0508, line 12: same here

templates/IAM/iam-role.yaml
```diff
@@ -0,0 +1,41 @@
...
+    Description: A list of managed policies for the role.
+  MaxSessionDuration:
+    Type: Number
+    Description: Maximum session duration in seconds
+    Default: 43200
+  RoleName:
```
zaro0508, line 18: in general, it's better to let AWS generate role names otherwise users might create conflicting names without knowing.  At the very least you should make this an optional parameter.

zaro0508 (changes requested):
ok, i see. you want to pass in a policy which is fine.  it seems like `cross-account-access.yaml` will do the same thing and create a policy managed policy as well.  Would that not work for your use case?

brucehoff:
> ok, i see. you want to pass in a policy which is fine. it seems like `cross-account-access.yaml` will do the same thing and create a policy managed policy as well. Would that not work for your use case?

I think I see what you are saying: The `PrincipalArns` parameter of `cross-account-access.yaml` goes into the "trust relationship" policy in the Role.  In my use case the principal is:

```
 "Service": "quicksight.amazonaws.com"
```

The comment in `cross-account-access.yaml` says that the parameter is a "list of accounts/users/roles".  Can it also be a *service*, like "quicksight.amazonaws.com"?

PR #394, opened 2023-06-20
IT-2736: Setup templates

xschildw:
This PR installs the templates needed to move from the existing github-oidc-provider.j2 to the new .njk template (with support for multiple repos.


templates/IAM/oidc-provider.yaml
```diff
@@ -0,0 +1,29 @@
...
+    Type: List<String>
+    Description: >-
+      A list of certificate thumbprints that are associated with the specified
+      IAM OIDC provider resource object
+  Url:
+    Type: String
+    Description: "The URL that the IAM OIDC provider resource object is associated with"
+Resources:
+  Provider:
+    Type: AWS::IAM::OIDCProvider
```
zaro0508, line 20: why do we need this template at all?  

also, instead of creating this wrapper why would you not just use this CFN resources directly?

templates/IAM/github-oidc-provider-access.njk
```diff
@@ -0,0 +1,115 @@
+{# Allow Github to access AWS without giving github AWS credentials               #}
```
zaro0508, line 1: now we have this file and `github-oidc-provider.j2`.  do we need both files?

templates/IAM/github-oidc-provider-access.njk
```diff
@@ -0,0 +1,115 @@
...
+        }
+  MaxSessionDuration:
+    Type: Number
+    Description: "The IAM role's maximum session duration"
+    MaxValue: 43200
+    MinValue: 3600
+    Default: 3600
+  GitHubOrg:
+      Type: String
+      Description: "The name of the github organization"
```
zaro0508, line 64: GitHubOrg is a jinja parameter, not a CFN parameter.  you can remove this.

templates/IAM/github-oidc-provider-access.njk
```diff
@@ -0,0 +1,115 @@
...
+{# Must pass in the following templating parameters                               #}
+{#  Parameters:                                                                   #}
+{#    ProviderArn: !CopyValue [ !Sub '${resourcePrefix}-${appName}-ProviderArn' ] #}
+{#    ManagedPolicyArns:                                                          #}
+{#      - "arn:aws:iam::aws:policy/AdministratorAccess"                           #}
+
+{# Make template compatible with both sceptre and org-formation     #}
+{% if sceptre_user_data.Repositories is defined %}
+    {% set Repositories = sceptre_user_data.Repositories %}
+{% endif %}
```
zaro0508, line 11: are we going to support this compatibility?  or do we only want this for sceptre?  if we want to support this then it needs to be tested and documented

zaro0508 (changes requested):

templates/IAM/github-oidc-provider-access.njk
```diff
@@ -0,0 +1,115 @@
+{# Allow Github to access AWS without giving github AWS credentials               #}
+{# Must pass in the following templating parameters                               #}
+{#  Parameters:                                                                   #}
```
zaro0508, line 3: we should probably add some comments similar to https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/org-formation/650-identity-providers/github-oidc-provider-access.njk#L1-L11

templates/IAM/github-oidc-provider-access.njk
zaro0508: GitHubOrg parameter has not been removed?

zaro0508 (changes requested):

templates/IAM/github-oidc-provider.j2
```diff
@@ -18,18 +17,13 @@ Parameters:
       A list of client IDs (also known as audiences) that are associated with
       the specified IAM OIDC provider resource object
     Default: "sts.amazonaws.com"
-  ThumbprintList:
```
xschildw, line 21: The provider is created in separate stack, so not needed anymore.

templates/IAM/github-oidc-provider.j2
```diff
@@ -1,12 +1,19 @@
...
+{#  Sceptre_user_data:                                                            #}
+{#    Repositories:                                                               #}
+{#     - name: "organizations-infra"                                              #}
+{#       branch: "master"                                                         #}
+{#    GitHubOrg: "Sage-Bionetworks"                                               #}
 
-{# Make template compatible with both sceptre and org-formation     #}
 {% if sceptre_user_data.Repositories is defined %}
     {% set Repositories = sceptre_user_data.Repositories %}
+    {% set GitHubOrg = sceptre_user_data.GitHubOrg %}
```
zaro0508, line 16: you'll need to put this in a separate if condition..

```
{% if sceptre_user_data.GitHubOrg is defined %}
    {% set GitHubOrg = sceptre_user_data.GitHubOrg %}
{% endif %}
```

zaro0508 (changes requested):

zaro0508 (approved):

PR #397, opened 2023-09-19
IT-3162: Support cert with altnames + update S3WebCloudfront

xschildw:
This PR creates a template to create a certificate with alternate subject names.
The existing template supports creating certs that cover *.<domainName> or xxx.domainName. 
Should I update existing instead?


xschildw:
The changes to S3WebCouldfront lock down the bucket to be accessed only through Cloudfront and other misc changes from recent versions.

xschildw:
Deployed both manually to create a test bucket and check the GH workflow.

templates/acm-certificate-alt-names.yaml
```diff
@@ -0,0 +1,38 @@
+# Setting up a certificate requires a manual step to enter the DNS Validation on godaddy.
```
zaro0508, line 1: is there an existing folder we can put this file in?  if not them maybe make a folder?

I don't understand the intention of this file.  is it meant to replace the existing acm-certificates.yaml file?  is it meant to provide additional functionality to acm-certifcates.yaml?  is it meant for a completely different use case than acm-certificates.yaml?

templates/acm-certificate-alt-names.yaml
```diff
@@ -0,0 +1,38 @@
...
+  HasValidationDomainName: !Not [!Equals [!Ref ValidationDomainName, '' ]]
+Resources:
+  Certificate:
+    Type: AWS::CertificateManager::Certificate
+    Properties:
+      DomainName: !Ref DomainName
+      SubjectAlternativeNames: !If [HasAlternativeNames, !Split [',', !Ref AlternativeNames], !Ref AWS::NoValue ]
+      DomainValidationOptions:
+        - DomainName: !Ref DomainName
+          ValidationDomain: !If [HasValidationDomainName, !Ref ValidationDomainName, !Ref DomainName]
```
zaro0508, line 28: since AlternativeNames and ValidateDomainName are optional would this even work if user decides to not provide values for those parameters?

templates/acm-certificate-alt-names.yaml
```diff
@@ -0,0 +1,38 @@
...
+      ValidationMethod: "DNS"
+      Tags:
```
zaro0508, line 30: tags are no longer needed in templates.  we should be deploying with cloudformation tags

templates/managed-s3WebCloudfront.yaml
```diff
@@ -18,67 +18,74 @@ Parameters:
     Description: The Amazon Resource Name (ARN) of an AWS Certificate Manager (ACM) certificate.
     AllowedPattern: "arn:aws:acm:.*"
     ConstraintDescription: must be a valid certificate ARN.
+Conditions:
+  IsProd: !Equals [!Ref SubDomainName, 'prod']
 Resources:
-  WebsiteLogBucket:
-    Type: 'AWS::S3::Bucket'
+  CloudFrontOAI:
+    Type: AWS::CloudFront::CloudFrontOriginAccessIdentity
```
zaro0508, line 25: this seems like a drastic change to this template.  what's the purpose?

templates/acm-certificate-alt-names.yaml
```diff
@@ -0,0 +1,38 @@
...
+    Default: ''
+Conditions:
+  HasAlternativeNames: !Not [!Equals [!Ref AlternativeNames, '' ]]
+  HasValidationDomainName: !Not [!Equals [!Ref ValidationDomainName, '' ]]
+Resources:
+  Certificate:
+    Type: AWS::CertificateManager::Certificate
+    Properties:
+      DomainName: !Ref DomainName
+      SubjectAlternativeNames: !If [HasAlternativeNames, !Split [',', !Ref AlternativeNames], !Ref AWS::NoValue ]
```
zaro0508, line 25: you don't have to use the split if you just changed the AlternativeNames parameter to accept a List instead of a string

zaro0508 (changes requested):
it's hard to tell how this is going to be used.  would you be able to craft a PR that shows the usage for these changes?

PR #398, opened 2023-11-16
PLFM-8005: Added stack parameter to original version in orginfra

xschildw:
This PR is a parameterized version of https://github.com/Sage-Bionetworks-IT/organizations-infra/blob/master/sceptre/synapseprod/templates/snowflake-access.yaml (added 'Stack' parameter).

### DIFF
--- a/templates/IAM/snowflake-synapse-access.yaml
+++ b/templates/IAM/snowflake-synapse-access.yaml
@@ -1,0 +1,68 @@
+Description: >
+  Setup cross account IAM access for Snowflake.
+  See https://docs.snowflake.com/en/user-guide/data-load-s3-config-storage-integration
+AWSTemplateFormatVersion: 2010-09-09
+Parameters:
+  Stack:
+    Type: String
+    AllowedValues:
+      - dev
+      - prod
+  SnowflakeAccountArn:
+    Type: String
+  SnowflakeAccountExternalId:
+    Type: String
+Resources:
+  SnowflakeServicePolicy:
+    Type: 'AWS::IAM::ManagedPolicy'
+    Properties:
+      PolicyDocument: |
+        {
+          "Version": "2012-10-17",
+          "Statement": [
+              {
+                  "Effect": "Allow",
+                  "Action": [
+                      "s3:GetObject",
+                      "s3:GetObjectVersion"
+                  ],
+                  "Resource": !Sub "arn:aws:s3:::${Stack}.datawarehouse.sagebase.org/warehouse/*"
+              },
+              {
+                  "Effect": "Allow",
+                  "Action": [
+                      "s3:ListBucket",
+                      "s3:GetBucketLocation"
+                  ],
+                  "Resource": !Sub "arn:aws:s3:::${Stack}.datawarehouse.sagebase.org",
+                  "Condition": {
+                      "StringLike": {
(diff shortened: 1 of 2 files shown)


zaro0508 (changes requested):
if the file in this PR serves the same purpose as the existing `snowflake-access.yaml` file  then why do you need to create a new file?  Can't you just update the existing file?

xschildw:
I'm moving it from sceptre/synapseprod/templates to aws-infra (and adding a param to support the stack) and will be using it for both synapseprod and synapsedev (synapsedev has a hardcoded test stack that is not in orginfra, we'll delete it and move everything to org-infra).

marcomarasca (commented):
Shouldn't we delete the existing snowflake yaml template in the synapse porod?

xschildw:
@marcomarasca , correct. The original file will be deleted from the organization-infra repo.

templates/IAM/snowflake-synapse-access.yaml
```diff
@@ -0,0 +1,60 @@
...
+          -
+            Effect: Allow
+            Action:
+                - s3:ListBucket
+                - s3:GetBucketLocation
+            Resource: !Sub "arn:aws:s3:::${Stack}.datawarehouse.sagebase.org"
+            Condition:
+              StringLike":
+                "s3:prefix":
+                  - "warehouse/*"
```
zaro0508, line 37:
```
            Condition:
              StringLike":
                "s3:prefix":
                  - !Sub "${BucketPrefix}/*"
```

templates/IAM/snowflake-synapse-access.yaml
```diff
@@ -0,0 +1,60 @@
...
+            Action:
+              - s3:GetObject
+              - s3:GetObjectVersion
+            Resource: !Sub "arn:aws:s3:::${Stack}.datawarehouse.sagebase.org/warehouse/*"
+          -
+            Effect: Allow
+            Action:
+                - s3:ListBucket
+                - s3:GetBucketLocation
+            Resource: !Sub "arn:aws:s3:::${Stack}.datawarehouse.sagebase.org"
```
zaro0508, line 33: to make it more generic i suggest you pass in a bucket ARN and prefix as parameters.
example:

```
Parameters:
  BucketArn:
    Type: String
  BucketPrefix:
    Type: String
```

for the PolicyDocument..
```
      PolicyDocument:
        Version: 2012-10-17
        Statement:
          -
            Effect: Allow
            Action:
              - s3:GetObject
              - s3:GetObjectVersion
            Resource: !Sub "${BucketArn}/${BucketPrefix}/*"
          -
            Effect: Allow
            Action:
                - s3:ListBucket
                - s3:GetBucketLocation
            Resource: !Sub "${BucketArn}"
```

zaro0508 (changes requested):
Ahh, i see. sorry i missed that the existing file is not in aws-infra.

templates/IAM/snowflake-synapse-access.yaml
xschildw: Great idea! Renamed since not tied to Synapse anymore. Thx!

zaro0508 (approved):

marcomarasca (approved):

PR #407, opened 2024-03-13
[IT-3471] templates to setup cloudwatch dashboard

zaro0508:
move files from org-formation[1] repo to here so it can be shared from multiple deployments.

[1] https://github.com/Sage-Bionetworks-IT/organizations-infra/tree/master/org-formation/740-cloudwatch-dashboard

### DIFF
--- a/.cfnlintrc.yaml
+++ b/.cfnlintrc.yaml
@@ -4,5 +4,6 @@ ignore_checks:
   - E3001
   - W2001
   - W3045
+  - W8001
 ignore_templates:
   - templates/tags/*.json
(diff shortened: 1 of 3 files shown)


.cfnlintrc.yaml
```diff
@@ -4,5 +4,6 @@ ignore_checks:
   - E3001
   - W2001
   - W3045
+  - W8001
```
zaro0508, line 7: ignores message `W8001 Condition DoCWReadOnly not used
templates/Cloudwatch/CrossAccountSharingRole-AccountList.yaml:29:3`

jesusaurus (approved):

PR #422, opened 2025-02-25
[IT-4144] CIS 5.1 Restrict source IP for SSH/RDP

ConsoleCatzirl:
The network ACL applies to the entire VPC, effecting both public and private subnets.

Modify the ACL to allow all traffic within the VPC and from the VPN, deny any other traffic to SSH or RDP, and allow all other traffic.

### DIFF
--- a/templates/vpc.yaml
+++ b/templates/vpc.yaml
@@ -280,6 +280,61 @@ Resources:
           Value: "Public"
         - Key: "Name"
           Value: "Public"
+  InboundAllowLocalNetworkAclEntry:
+    Type: "AWS::EC2::NetworkAclEntry"
+    Properties:
+      NetworkAclId:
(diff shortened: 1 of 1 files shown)


zaro0508 (commented):
There are already existing ACL resource, InboundHTTPPublicNetworkAclEntry. should we remove that?

ConsoleCatzirl:
@zaro0508 no, we want to keep that entry and also add two more before it: the first rule allows all traffic from the VPN, the second rule blocks all other SSH and RDP traffic (meaning it didn't come from the VPN), and finally we fall back on the existing rule to allow all other traffic.

zaro0508:
please take a look at PR https://github.com/Sage-Bionetworks-IT/organizations-infra/pull/1343

PR #424, opened 2025-03-24
[IT-2361] Fix nextflow batch compute env template

zaro0508:
* Fix launch template reference.  It can only take one of name or id.
* Fix batch role to resolve the following stabalization failure:
```
Resource handler returned message: "Resource of type
'AWS::Batch::ComputeEnvironment' with identifier
'arn:aws:batch:us-east-1:035458030717:compute-environment/ComputeEnvironmentSpot-eULJVaK00uH15ZfV'
did not stabilize."
```

### DIFF
--- a/templates/NextflowTower/nextflow-launch.yaml
+++ b/templates/NextflowTower/nextflow-launch.yaml
@@ -7,12 +7,24 @@ Parameters:
   SubnetIds:
     Type: CommaDelimitedList
     Description: List of public Subnet Ids for the load balancer
+  MinvCpus:
+    Type: Number
+    Description: The number of min CPUs for Batch compute environments
+    MinValue: 0
+    MaxValue: 16
+    Default: 0
   MaxvCpus:
     Type: Number
     Description: The number of max CPUs for Batch compute environments
-    MinValue: 1
-    MaxValue: 16
-    Default: 1
+    MinValue: 8
+    MaxValue: 64
+    Default: 8
+  DesiredvCpus:
+    Type: Number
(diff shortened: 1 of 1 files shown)


templates/NextflowTower/nextflow-launch.yaml
```diff
@@ -10,15 +10,21 @@ Parameters:
   MinvCpus:
     Type: Number
     Description: The number of min CPUs for Batch compute environments
-    MinValue: 1
+    MinValue: 0
     MaxValue: 16
-    Default: 1
+    Default: 0
```
BWMac, line 15: Does this mean that all Nextflow processes will need to have a number of CPUs specified?

zaro0508: This is a scaling configuration which, i believe, will allow batch to scale down to 0 resources when no processes are assigned to this batch pipeline.   

As I understand it,  you don't configure this as a tower end user.   This is a shared batch environment,  Tower should allow you to assign your processes to this batch pipeline.

BWMac: When running a Nextflow process, you can set the number of CPUs and amount of memory to be used for that specific process https://www.nextflow.io/docs/latest/reference/process.html#process-cpus. In our current deployment, when you do not specify a number of CPUs it defaults to 1, but this change seems to make that default 0. I am wondering what the impact of that change will be when one does not specify the cpu directive in a process definition.

BWMac (approved):

PR #425, opened 2025-04-05
PLFM-8908 template to add CIS Level 1 image image builder for Synapse

brucehoff:
This PR adds a template to create an image builder pipeline to build a CIS Level 1 hardened image which will work with Elastic Beanstalk, for use by Synapse.

### DIFF
--- a/templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
+++ b/templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
@@ -1,0 +1,138 @@
+---
+AWSTemplateFormatVersion: 2010-09-09
+Parameters:
+  AwsOrganizationId:
+    Description: Share the generated image with this Organization
+    Type: String
+    Default: "o-69lcdj4kro"
+  ImageVersion:
+    Description: The generated image version.
+    Type: String
+    Default: "0.0.0"
+  VolumeSize:
+    Description: The EBS volume size (in GB)
+    Type: Number
+    Default: 50
+    MinValue: 8
+    MaxValue: 500
+Resources:
+  # By default, AWS Services do not have permission to perform actions on your instances. This grants
+  # AWS Systems Manager (SSM) and EC2 Image Builder the necessary permissions to build an image.
+  # https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-resource-iam-role.html
+  # https://docs.aws.amazon.com/imagebuilder/latest/userguide/image-builder-setting-up.html
+  InstanceRole:
+    Type: AWS::IAM::Role
+    Metadata:
+      Comment: Role to be used by instance during image build.
+    Properties:
+      ManagedPolicyArns:
+        - Fn::Sub: arn:${AWS::Partition}:iam::aws:policy/AmazonSSMManagedInstanceCore
+        - Fn::Sub: arn:${AWS::Partition}:iam::aws:policy/EC2InstanceProfileForImageBuilder
+      AssumeRolePolicyDocument:
+        Statement:
+          - Action:
+              - sts:AssumeRole
+            Effect: Allow
+            Principal:
+              Service:
+                - !Sub 'ec2.${AWS::URLSuffix}'
+        Version: '2012-10-17'
+      Path: /executionServiceEC2Role/
+  # To pass the InstanceRole to an EC2 instance, we need an InstanceProfile.
+  # This profile will be used during the image build process.
+  # https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_use_switch-role-ec2_instance-profiles.html
+  InstanceProfile:
+    Type: AWS::IAM::InstanceProfile
+    Properties:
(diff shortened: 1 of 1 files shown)


xschildw (approved):
LGTM, cross-checked against amazon-linux-2023-agora-bastian.yaml + docs.

templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
```diff
@@ -0,0 +1,138 @@
...
+          DeleteOnTermination: true
+        DeviceName: "/dev/xvda"
+      AdditionalInstanceConfiguration:
+        SystemsManagerAgent:
+          UninstallAfterBuild: false
+
+  CISforEBImageBuilderComponent:
+    Type: AWS::ImageBuilder::Component
+    Properties:
+      Name: CISforEBImageBuilderComponent
```
zaro0508, line 98: it's usually not good to have a static name, to reduce conflict i suggest you set this to `${StackName}` or prepend it to this value.

templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
```diff
@@ -0,0 +1,138 @@
...
+        - ComponentArn:
+            Ref: CISforEBImageBuilderComponent
+      # The image will not build with the default working dir, /tmp, since CIS makes that directory non-executable
+      WorkingDirectory: "/root"
+      BlockDeviceMappings:
+      - Ebs:
+          Throughput: 125
+          VolumeType: gp3
+          Iops: 3000
+          VolumeSize: !Ref VolumeSize
```
zaro0508, line 88: you probably only need to specify `VolumeSize` parameter.  The other ones are not needed because i believe they are just the default values.

templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
```diff
@@ -0,0 +1,138 @@
...
+          DeleteOnTermination: true
+        DeviceName: "/dev/xvda"
+      AdditionalInstanceConfiguration:
+        SystemsManagerAgent:
+          UninstallAfterBuild: false
```
zaro0508, line 93: you don't need to specify this because the default is `false`

templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
```diff
@@ -0,0 +1,138 @@
...
+        - Region: !Ref 'AWS::Region'
+          AmiDistributionConfiguration:
+            AmiTags:    # apply tags to generated AMIs
+              Name: !Sub '${AWS::StackName}'
+            LaunchPermissionConfiguration:
+              OrganizationArns:
+                - !Sub 'arn:${AWS::Partition}:organizations::${AWS::AccountId}:organization/${AwsOrganizationId}'
+              UserIds:
+                - 449435941126   # Synapse dev
+                - 325565585839   # Synapse prod
```
zaro0508, line 138: why not share to all accounts by setting to `- all`?

templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
```diff
@@ -0,0 +1,138 @@
...
+        Ref: ImageDistributionConfiguration
+      ImageTestsConfiguration:
+        TimeoutMinutes: 720
+        ImageTestsEnabled: true
+
+  Recipe:
+    Type: "AWS::ImageBuilder::ImageRecipe"
+    Properties:
+      Name: !Sub ${AWS::StackName}
+      Version: !Ref ImageVersion
```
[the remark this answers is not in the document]
brucehoff: > you have a pipeline that builds every 7 days with the same image version? Does this even work?  Doesn't the version number need to be unique on every build?
Here `Version` is the version of the Recipe, not the version of the image.  The point is to build using the same Recipe each time.  The Recipe uses the latest version of the CIS Level 1 image.  Therefore the build pipeline will automatically pick up patches/updates provided by CIS.

> Also why would you need to build every seven days when you can just bump this version number to trigger a new build?
Hopefully the explanation above answers this question.

templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
brucehoff: Since the base image is purchased from the AWS Marketplace, setting to 'all' results in an error.

zaro0508: hmm, is that limitation documented anywhere?  also did you try `UserGroups: -all` parameter, not `UserIds:`

brucehoff: > hmm, is that limitation documented anywhere? 
Not that I've seen.

> also did you try `UserGroups: -all` parameter, not `UserIds:`
Yes, I originally tried following the pattern you used in your Agora template, but it failed, as described above.

PR #426, opened 2025-04-09
[IT-4269] Add a template for a data transfer budget

ConsoleCatzirl:
Duplicate the budget template, adding a cost filter for data-transfer usage types. The list of usage types was found using Cost Explorer. The format of the usage type string was verified by manually creating a budget in AWS and using `aws budgets describe-budgets` to view the auto-generated cost filter.

Docs:
https://docs.aws.amazon.com/AWSCloudFormation/latest/UserGuide/aws-properties-budgets-budget-budgetdata.html https://docs.aws.amazon.com/cost-management/latest/userguide/budgets-create-filters.html https://docs.aws.amazon.com/cli/latest/reference/budgets/describe-budgets.html

### DIFF
--- a/templates/Billing/budgets-transfer-costs.yaml
+++ b/templates/Billing/budgets-transfer-costs.yaml
@@ -1,0 +1,118 @@
+AWSTemplateFormatVersion: '2010-09-09'
+
+Parameters:
+
+  costAuditEmail:
+    Type: String
+    Default: 'cloud-cost-audit@sagebase.org'
+
+  synapseAdminEmail:
+    Type: String
+    Default: 'platform@sagebase.org'
+
+  techTeamEmail:
+    Type: String
+    Default: 'tech-team@sagebase.org'
+
+  budgetAmount:
+    Type: Number
+
+Resources:
+
+  NotificationTopic:
+    Type: AWS::SNS::Topic
+    Properties:
+      DisplayName: Data Transfer Budget Notifications
+
+  costAuditSubscription:
+    Type: AWS::SNS::Subscription
+    Properties:
+      Endpoint: !Ref costAuditEmail
+      Protocol: email
+      TopicArn: !Ref NotificationTopic
+
+  synapseAdminSubscription:
+    Type: AWS::SNS::Subscription
+    Properties:
+      Endpoint: !Ref synapseAdminEmail
+      Protocol: email
+      TopicArn: !Ref NotificationTopic
+
+  techTeamSubscription:
+    Type: AWS::SNS::Subscription
+    Properties:
+      Endpoint: !Ref techTeamEmail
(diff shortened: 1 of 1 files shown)


templates/Billing/budgets-transfer-costs.yaml
```diff
@@ -0,0 +1,104 @@
...
+
+  hasSubscription: !Not [ !Equals [ '', !Ref emailAddress ] ]
+
+Resources:
+
+  NotificationTopic:
+    Type: AWS::SNS::Topic
+    Properties:
+      DisplayName: Budgets Notification
+      TopicName: !Sub '${budgetName}-topic'
```
zaro0508, line 25: This does not seem like a required property.  I would recommend letting AWS manage the topic name

templates/Billing/budgets-transfer-costs.yaml
```diff
@@ -0,0 +1,104 @@
...
+            - "USE1-APE1-AWS-Out-Bytes"
+            - "USE1-APN1-AWS-Out-Bytes"
+            - "USE1-APS1-AWS-Out-Bytes"
+            - "USE1-APS3-AWS-Out-Bytes"
+            - "USE1-EUW2-AWS-Out-Bytes"
+            - "USE1-USE2-AWS-Out-Bytes"
+            - "USE1-USW1-AWS-Out-Bytes"
+            - "USE1-USW2-AWS-Out-Bytes"
+      NotificationsWithSubscribers:
+        - Notification:
```
zaro0508, line 81: seems spam induced with all these notifications?

zaro0508 (changes requested):

templates/Billing/budgets-transfer-costs.yaml
ConsoleCatzirl: To be honest I didn't think too much about them when I made the copy of `templates/Billing/budgets.yaml`, they seemed reasonable at first glance. I see you wrote the original template a few years ago, has the number of notifications for the current budgets been an issue?

zaro0508: The only place where I see `budgets.yaml` being used is to setup notifications for general account budgets.   TBH, the only time where I've seen alerts are from accounts for scientific projects which I ignore because i have not idea whether it's warranted.  Although I think they are already kinda spammy.  I guess i don't i don't mind piling but just wanted to point it out.

zaro0508 (approved):

templates/Billing/budgets-transfer-costs.yaml
ConsoleCatzirl: ah ok, thanks for pointing it out. I think we can start here and adjust when we need to

PR #432, opened 2025-05-20
CIS Image pipeline: Adding changes for CIS compliance scan findings

brucehoff:
CIS-4.2.6  Ensure SSH PAM is enabled
CIS-4.2.18 Ensure SSH MaxSessions is set to 10 or less
CIS-4.6.6  Ensure root password is set
CIS-4.2.20 Ensure SSH Idle Timeout Interval is configured

### DIFF
--- a/templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
+++ b/templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
@@ -111,18 +111,43 @@ Resources:
                 inputs:
                   commands:
                     - sudo /usr/sbin/useradd --user-group xray -s /sbin/nologin --no-create-home
-              - name: CIS-5114
+              - name: CIS-5_1_14
                 # Edit ... /etc/rsyslog.conf ... and set $FileCreateMode to 0640 or more restrictive: $FileCreateMode 0640
                 action: ExecuteBash
                 inputs:
                   commands:
                     - sudo echo "\\$FileCreateMode 0640" >> /etc/rsyslog.conf
-              - name: CIS-5113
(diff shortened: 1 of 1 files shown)


templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
```diff
@@ -111,18 +111,43 @@ Resources:
...
                 action: ExecuteBash
                 inputs:
                   commands:
                     - sudo sed -i "s|#ForwardToSyslog=no|ForwardToSyslog=yes|g" /etc/systemd/journald.conf
+              - name: CIS-4_2_20
+                # Ensure SSH Idle Timeout Interval is configured
+                action: ExecuteBash
+                inputs:
+                  commands:
+                    - sudo sed -i "s|ClientAliveInterval 0|ClientAliveInterval 300|g" /etc/ssh/sshd_config
```
ConsoleCatzirl, line 131: This works as-is, but since the left-hand side is a regular expression we also have the ability future-proof against changes to default values with a wildcard match, e.g.:

`sed -i "s|ClientAliveInterval .*|ClientAliveInterval 300|" /etc/ssh/sshd_config`

ConsoleCatzirl (changes requested):

templates/ImageBuilder/cis-for-eb-image-pipeline-template.yaml
ConsoleCatzirl: ah, I see, so in this particular case it doesn't matter.

ConsoleCatzirl (approved):

PR #436, opened 2025-08-27
[IT-4498] Template to setup redirect

zaro0508:
Add a template to setup a redirect from one apex domain to another
For example: redirect all traffic from my-site.org to mysite.org

### DIFF
--- a/templates/S3/s3-apex-redirector.yaml
+++ b/templates/S3/s3-apex-redirector.yaml
@@ -1,0 +1,69 @@
+# Setup a redirect from one apex domain to another
+# For example: redirect all traffic from my-site.org to mysite.org
+# This template is based on https://repost.aws/knowledge-center/redirect-domain-route-53
+# It currently only works in us-east-1 region
+AWSTemplateFormatVersion: 2010-09-09
+Description: Setup redirect from one apex domain to another using S3 static website redirect
+Parameters:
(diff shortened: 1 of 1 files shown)


xschildw:
I think you need Cloudfront on the source domain to provide https (with a cert).

zaro0508:
> I think you need Cloudfront on the source domain to provide https (with a cert).

Why?  The AWS blog post says (for the simple case) just use an S3 bucket and configure it for redirect.  There are no steps for cloudfront.

xschildw:
Without Cloudfront, you can't setup a certificate and have https. See the note in the blogpost 'Amazon S3 website endpoints don't support HTTPS. To redirect both HTTP and HTTPS requests, use other methods such as an Application Load Balancer or Amazon CloudFront with an edge function.' . I'm arguing for the latter solution (with no bucket).

zaro0508:
> Without Cloudfront, you can't setup a certificate and have https.
This solution does not use a certificate at all.  `treat-ad.org` just redirects to `treatad.org`.  `treatad.org` is the one that uses a https certificate.  i have deployed this solution and it seems to work for me, treat-ad.org is redirect to treatad.org.

zaro0508:
This solution only works for http, not https.  not good enough.

xschildw (approved):
LGTM

PR #444, opened 2026-04-09
IT-4552: Use OAC to restrict access to the bucket

xschildw:
This PR updates the template to deploy an S3 bucket and associated distribution to match latest versions deployed as part of moving them into the infra.
Main change: restrict access to bucket using OAC

### DIFF
--- a/templates/managed-s3WebCloudfront.yaml
+++ b/templates/managed-s3WebCloudfront.yaml
@@ -18,8 +18,16 @@ Parameters:
     Description: The Amazon Resource Name (ARN) of an AWS Certificate Manager (ACM) certificate.
     AllowedPattern: "arn:aws:acm:.*"
     ConstraintDescription: must be a valid certificate ARN.
+  OnErrorGotoRoot:
+    Type: String
+    Default: false
+    AllowedValues:
+      - true
+      - false
+    Description: Set to true to redirect 403/404 errors to /index.html (SPA routing)
 Conditions:
   IsProd: !Equals [!Ref SubDomainName, 'prod']
+  UseCustomErrorResponses: !Equals [!Ref OnErrorGotoRoot, 'true']
 Resources:
   AccessLogBucket:
     Type: AWS::S3::Bucket
@@ -105,15 +113,17 @@ Resources:
         Aliases:
           - !Sub '${SubDomainName}.${DomainName}'
           - !If [IsProd, !Sub '${DomainName}', !Ref AWS::NoValue]
-        CustomErrorResponses:
-          - ErrorCachingMinTTL: 60
-            ErrorCode: 404
-            ResponseCode: 200
-            ResponsePagePath: '/index.html'
-          - ErrorCachingMinTTL: 60
-            ErrorCode: 403
-            ResponseCode: 200
-            ResponsePagePath: '/index.html'
+        CustomErrorResponses: !If
+          - UseCustomErrorResponses
+          - - ErrorCachingMinTTL: 60
+              ErrorCode: 404
+              ResponseCode: 200
+              ResponsePagePath: '/index.html'
+            - ErrorCachingMinTTL: 60
+              ErrorCode: 403
+              ResponseCode: 200
+              ResponsePagePath: '/index.html'
+          - !Ref "AWS::NoValue"
         DefaultCacheBehavior:
           DefaultTTL: 3600
           AllowedMethods:
--- a/templates/managed-s3WebCloudfront.yaml
+++ b/templates/managed-s3WebCloudfront.yaml
@@ -18,10 +18,6 @@ Parameters:
     Description: The Amazon Resource Name (ARN) of an AWS Certificate Manager (ACM) certificate.
     AllowedPattern: "arn:aws:acm:.*"
     ConstraintDescription: must be a valid certificate ARN.
-  LogPrefix:
-    Type: String
-    Default: "cloudfront-logs/"
-    Description: Prefix for log files in the log bucket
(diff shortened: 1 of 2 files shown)


templates/managed-s3WebCloudfront.yaml
```diff
@@ -18,108 +18,136 @@ Parameters:
...
+          - ObjectOwnership: BucketOwnerEnforced
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
       LoggingConfiguration:
-        DestinationBucketName: !Ref WebsiteLogBucket
+        DestinationBucketName: !Ref AccessLogBucket
+        LogFilePrefix: !Sub 's3-access/${AWS::StackName}/'
```
zaro0508, line 79: string containing the `LogPrefix` input parameter or AWS::StackName ?

templates/managed-s3WebCloudfront.yaml
```diff
@@ -18,108 +18,136 @@ Parameters:
...
+      PublicAccessBlockConfiguration:
+        BlockPublicAcls: true
+        BlockPublicPolicy: true
+        IgnorePublicAcls: true
+        RestrictPublicBuckets: true
+      LifecycleConfiguration:
+        Rules:
+          - Status: Enabled
+            ExpirationInDays: 365
+            Prefix: !Ref LogPrefix
```
zaro0508: why not just use `AWS::StackName`?

templates/managed-s3WebCloudfront.yaml
```diff
@@ -18,108 +18,136 @@ Parameters:
...
+          - !If [IsProd, !Sub '${DomainName}', !Ref AWS::NoValue]
+        CustomErrorResponses:
+          - ErrorCachingMinTTL: 60
+            ErrorCode: 404
+            ResponseCode: 200
+            ResponsePagePath: '/index.html'
+          - ErrorCachingMinTTL: 60
+            ErrorCode: 403
+            ResponseCode: 200
+            ResponsePagePath: '/index.html'
```
zaro0508: why are we returning 200 for errors?

templates/managed-s3WebCloudfront.yaml
```diff
@@ -18,108 +18,136 @@ Parameters:
...
           ViewerProtocolPolicy: redirect-to-https
-        PriceClass: PriceClass_All
+        PriceClass: PriceClass_100
         ViewerCertificate:
           AcmCertificateArn: !Ref AcmCertificateArn
+          MinimumProtocolVersion: TLSv1.2_2021
           SslSupportMethod: sni-only
+        Logging:
+          Bucket: !GetAtt AccessLogBucket.DomainName
+          Prefix: !Ref LogPrefix
```
zaro0508: same here, string with `AWS::StackName` or `LogPrefix`?

zaro0508 (changes requested):

templates/managed-s3WebCloudfront.yaml
xschildw: That's the way it was requested for portals, should I make it optional ?

templates/managed-s3WebCloudfront.yaml
xschildw: Done

templates/managed-s3WebCloudfront.yaml
xschildw: Agree

templates/managed-s3WebCloudfront.yaml
xschildw: Made it optional

templates/managed-s3WebCloudfront.yaml
xschildw: Done

templates/managed-s3WebCloudfront.yaml
zaro0508: I have no idea if making it optional makes anymore sense.  For me to determine that I would need to understand your intentions.  I looked at the jira issue to see if there's anymore info however it looks like the issue you reference (IT-4552) doesn't even align with this PR?

xschildw: Most of the static websites we deploy are SPA (React) applications (where the pattern of returning index.html on 403/404 is used). You rightfully pointed that out and indeed there are cases when we just want to return the error, hence making the behavior optional. I can rename the option to 'SpaErrorHandling' and add some comment, or would you prefer separate scripts for SPA and regular?

zaro0508: how is this PR related to issue IT-4552?  isn' t-4552 about documentation?

xschildw: From the issue description 'The existing sites have been created using a variety of scripts, there should also be scripts in the aws-infra account. **We should update these scripts** and document the whole process (create or reference certificates, bucket, distributions, DNS) so that future sites are created in a consistent way.'. 
This PR updates the script used to deploy the bucket and the distribution, given a domain/subdomain, cert arn (and now error handling behavior). Should I split the update script part off original issue?

zaro0508: ok, i was confused because this PR does not update scripts nor does it update any docs.  maybe you meant script as in cloudformation template?  i don't really consider a yaml file a script.  It would probably be a good idea to change the wording and provide more detail in the jira issue.

zaro0508 (approved):

